PR #1, opened 2019-05-17
CIS scan

leodotcloud:


cmd/kube-bench-summarizer/main.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+}
+
+func run(c *cli.Context) error {
+	logrus.Info("Running Summarizer")
+	inputDir := c.String(INPUT_DIR_FLAG)
+	outputDir := c.String(OUTPUT_DIR_FLAG)
+
+	if inputDir == "" {
+		logrus.Errorf("error: %v not specified", INPUT_DIR_FLAG)
```
alena1108: any reason of why we don't return an error?

cmd/kube-bench-summarizer/main.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil
+	}
+
+	if outputDir == "" {
+		logrus.Errorf("error: %v not specified", OUTPUT_DIR_FLAG)
+		return nil
```
alena1108: same question here

cmd/kube-bench-summarizer/main.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	s := summarizer.NewSummarizer(
+		inputDir,
+		outputDir,
+	)
+	if err := s.Summarize(); err != nil {
+		logrus.Fatalf("error summarizing: %v", err)
```
alena1108: Could you describe the diff of where it would make sense to fatal, and where returning an error is fine?

cmd/kube-bench-summarizer/summarizer/summarizer.go
```diff
@@ -0,0 +1,200 @@
...
+
+	"encoding/json"
+	"fmt"
+	kb "github.com/aquasecurity/kube-bench/check"
+	"os/exec"
+	"path/filepath"
+)
+
+const (
+	DEFAULT_OUTPUT_FILE_NAME = "report.json"
```
alena1108: can you think of a use case where we want to parameterize this param? Given its name starts with DEFAULT, assuming that we consider it could be changed.

cmd/kube-bench-summarizer/summarizer/summarizer.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		return err
+	}
+
+	for _, fileName := range resultFiles {
+		// Load one result file
+		// Marshal it into the results
+		contents, err := ioutil.ReadFile(filepath.Clean(fileName))
+		if err != nil {
+			logrus.Errorf("error reading file %+v: %v", fileName, err)
```
alena1108: Nit: it's better for the method to just return the error, and get it logged in the caller stack.

cmd/kube-bench-summarizer/summarizer/summarizer.go
```diff
@@ -0,0 +1,200 @@
...
+		err = json.Unmarshal(contents, results)
+		if err != nil {
+			logrus.Errorf("error unmarshalling: %v", err)
+			return err
+		}
+		logrus.Debugf("results: %+v", results)
+
+		for _, group := range results.Groups {
+			for _, check := range group.Checks {
+				logrus.Infof("id: %v %v", check.ID, check.State)
```
alena1108: nit: can we extract the code block inside the for loop into its own method? This way it would be easier to read

cmd/kube-bench-summarizer/summarizer/summarizer.go
```diff
@@ -0,0 +1,200 @@
...
+	logrus.Debugf("summarizeForHost: %v", hostname)
+
+	hostDir := fmt.Sprintf("%v/%v", s.InputDirectory, hostname)
+	resultFiles, err := filepath.Glob(fmt.Sprintf("%v/*.json", hostDir))
+	if err != nil {
+		return err
+	}
+
+	for _, fileName := range resultFiles {
+		// Load one result file
```
alena1108: Nit: it's better for the method to just return the error, and get it logged in the caller stack.

alena1108: So one method would be for readFile, another for readGroup

cmd/kube-bench-summarizer/summarizer/summarizer.go
```diff
@@ -0,0 +1,200 @@
...
+						switch check.State {
+						case kb.FAIL:
+							g.Fail++
+							s.summarizedReport.Fail++
+						case kb.WARN:
+							g.Warn++
+							s.summarizedReport.Warn++
+						case kb.INFO:
+							g.Info++
+							s.summarizedReport.Info++
```
alena1108: shall we fall back to some default?

cmd/kube-bench-summarizer/main.go
leodotcloud: will fix

cmd/kube-bench-summarizer/main.go
leodotcloud: will fix

cmd/kube-bench-summarizer/main.go
leodotcloud: I think I changed it to Fatal so that I can get a non-zero return code, will check again

cmd/kube-bench-summarizer/summarizer/summarizer.go
leodotcloud: The file name is being passed here: https://github.com/rancher/cis-k8s/pull/1/files#diff-95f2857257f0727f64e76eadc2945514R33 ... in case we change it in future or use some other tool.

cmd/kube-bench-summarizer/summarizer/summarizer.go
leodotcloud: 👍

cmd/kube-bench-summarizer/summarizer/summarizer.go
leodotcloud: 👍

cmd/kube-bench-summarizer/summarizer/summarizer.go
leodotcloud: There is no other state, but adding a default doesn't hurt.

cmd/kube-bench-summarizer/summarizer/summarizer.go
```diff
@@ -0,0 +1,200 @@
...
+
+func (s *Summarizer) printSummaryReport() error {
+	bytes, err := json.Marshal(s.summarizedReport)
+	if err != nil {
+		logrus.Errorf("error marshalling summary report: %v", err)
+		return err
+	}
+
+	txt := string(bytes)
+	logrus.Infof("txt: %+v", txt)
```
alena1108: do we intent to print it in the log? or was it more for debug purpose (in this case , we should reduce the log level to debug)

cmd/kube-bench-summarizer/summarizer/summarizer.go
```diff
@@ -0,0 +1,200 @@
...
+
+	for _, hostDir := range hostsDir {
+		if !hostDir.IsDir() {
+			continue
+		}
+		hostname := hostDir.Name()
+		logrus.Debugf("hostDir: %v", hostname)
+
+		if err := s.summarizeForHost(hostname); err != nil {
+			logrus.Errorf("error summarizeForHost: %v", hostname)
```
alena1108: do we want to just log the error and proceed? I am thinking of a case when the summarizer would fail, but rancher gets no error back.

package/run.sh
```diff
@@ -0,0 +1,56 @@
...
+KBS_OUTPUT_DIR=${KUBE_BENCH_SUMMARIZER_ROOT}/output
+KBS_OUPTPUT_FILENAME=report.json
+
+kube-bench-summarizer --input-dir ${KBS_INPUT_DIR} --output-dir ${KBS_OUTPUT_DIR}
+if [ $? -ne 0 ]; then
+  echo "error running kube-bench-summarizer"
+  exit 1
+fi
+
+# Create a config map with results
```
alena1108: is there a reason why we can't save the result to the config map from the go code directly?

package/run.sh
```diff
@@ -0,0 +1,56 @@
...
+kubectl -n ${SONOBUOY_NS} \
+  create cm cis-$(date +"%Y-%m-%d-%H-%M-%S-%N") \
+  --from-file ${KBS_OUTPUT_DIR}/${KBS_OUPTPUT_FILENAME}
+if [ $? -ne 0 ]; then
+  echo "error creating configmap for storing the report"
+  exit 1
+fi
+
+# Annotate self (pod) to signal "done"
+kubectl -n ${SONOBUOY_NS} \
```
alena1108: Do we intend to ever cleanup older config maps? as we don't really need more than one entry - at least not for this release - plus they can occupy etcd. Also we might want to think of how easy it would be to find the config map of the last run easily.

package/run.sh
```diff
@@ -0,0 +1,56 @@
...
+  annotate pod ${SONOBUOY_POD_NAME} \
+  ${DONE_ANNOTATION_KEY}=${DONE_ANNOTATION_VALUE}
+
+# Wait
+sleep infinity
```
alena1108: what eventually kills the pod/process?

cmd/kube-bench-summarizer/summarizer/summarizer.go
leodotcloud: Agree with you, I think we should return an error so that the problem can be debugged.

package/run.sh
leodotcloud: The summarizer is just summarizing the results, adding the go code to create the configmap isn't the right.

package/run.sh
leodotcloud: The cleanup happens with the CRD deletion.

package/run.sh
leodotcloud: The DaemonSet is cleaned up by sonobuoy. And sonobuoy itself will be cleaned up by the controller.

alena1108:
LGTM

PR #9, opened 2019-12-23
fix mismatching checks

leodotcloud:
- The IDs for a couple of checks were swapped incorrectly, fixed.
- Some of the scored tests related to PSP were marked unscored incorrectly in upstream, fixed.

Related upstream PR: https://github.com/aquasecurity/kube-bench/pull/544

### DIFF
--- a/package/cfg/cis-1.4/master.yaml
+++ b/package/cfg/cis-1.4/master.yaml
@@ -1506,7 +1506,7 @@ groups:
     remediation: |
      [Manual test]
      Create a PSP as described in the Kubernetes documentation, ensuring that the .spec.hostPID field is omitted or set to false.
-    scored: false
+    scored: true
 
   - id: 1.7.3
     text: "Do not admit containers wishing to share the host IPC namespace (Scored)"
@@ -1514,7 +1514,7 @@ groups:
     remediation: |
       [Manual test]
       Create a PSP as described in the Kubernetes documentation, ensuring that the .spec.hostIPC field is omitted or set to false.
-    scored: false
+    scored: true
 
   - id: 1.7.4
     text: "Do not admit containers wishing to share the host network namespace (Scored)"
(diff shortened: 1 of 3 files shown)


leodotcloud:
@prachidamle It's not Scored.

prachidamle:
@leodotcloud so do we just copy over the "scored" tests from the upstream and run those as part of the scan?

leodotcloud:
@prachidamle we have a copy of the upstream tests in `cis-1.4` folder and we keep only the Scored ones with the necessary modifications in rke-cis-1.4.

package/cfg/cis-1.4/master.yaml
```diff
@@ -1501,31 +1501,31 @@ groups:
     scored: false
 
   - id: 1.7.2
-    text: "Do not admit containers wishing to share the host process ID namespace (Not Scored)"
+    text: "Do not admit containers wishing to share the host process ID namespace (Scored)"
     type: "manual"
     remediation: |
      [Manual test]
      Create a PSP as described in the Kubernetes documentation, ensuring that the .spec.hostPID field is omitted or set to false.
     scored: false
```
prachidamle: Set `scored: true` for these tests?

package/cfg/cis-1.4/master.yaml
```diff
@@ -1501,31 +1501,31 @@ groups:
...
 
   - id: 1.7.4
-    text: "Do not admit containers wishing to share the host network namespace (Not Scored)"
+    text: "Do not admit containers wishing to share the host network namespace (Scored)"
     type: "manual"
     remediation: |
       [Manual test]
       Create a PSP as described in the Kubernetes documentation, ensuring that the .spec.hostNetwork field is omitted or set to false.
     scored: false
```
prachidamle: Same as above for all the scored PSP tests

package/cfg/cis-1.4/master.yaml
leodotcloud: good catch, I had fixed them for upstream, missed on local. Will fix.

package/cfg/cis-1.4/master.yaml
leodotcloud: good catch, I had fixed them for upstream, missed on local. Will fix.

PR #18, opened 2020-07-07
grep for kube-apiserver

prachidamle:
https://github.com/rancher/rancher/issues/26598

Changing the pgrep to use "kube-apiserver"


dramich:
LGTM but based on @leodotcloud comment in the issue this is probably not the end of having to massage the greps.

PR #23, opened 2020-08-17
Changes for cis v2

prachidamle:
- Bake in the KDM rke 20 NA and 14 skip tests into the yaml files of the benchmark using type="skip"
- Change report summarizer to return NA for these tests
- Change image building to download upstream kube-bench eks-1.0 and gke-1.0 benchmarks and change the config.yaml to support these benchmarks

https://github.com/rancher/cis-operator/issues/9
https://github.com/rancher/cis-operator/issues/7
https://github.com/rancher/cis-operator/issues/24


prachidamle:
Need this PR for the CIS v2 changes.

StrongMonkey:
@prachidamle Don't have enough domain knowledge for this, this looks OK to me.

package/Dockerfile
```diff
@@ -1,5 +1,15 @@
...
+FROM ubuntu:18.04 as intermediate
+
+#Label the image for cleaning after build process
+LABEL stage=intermediate
+
+RUN apt-get update \
+    && apt-get install -y \
+        git
+
+RUN git clone https://github.com/aquasecurity/kube-bench.git
```
briandowns, line 10: Do we need the entire history? If not, can we update to add `--depth=1`?

package/cfg/config.yaml
```diff
@@ -202,3 +212,15 @@ target_mapping:
...
+    - "node"
+    - "controlplane"
+    - "etcd"
+    - "policies"
+  "eks-1.0":
+    - "node"
+  "gke-1.0":
+    - "node"
+    - "managedservices"
+    - "policies"
```
briandowns, line 226: New line 

```suggestion
    - "policies"

```

package/Dockerfile
leodotcloud: Also, can we lock down the release tag as well?

pkg/kb-summarizer/summarizer/summarizer.go
```diff
@@ -249,6 +249,9 @@ func (s *Summarizer) processOneResultFileForHost(results *kb.Controls, hostname
 				continue
 			}
 
+			if check.Type == CheckTypeSkip {
+				check.State = NA
```
leodotcloud, line 253: If the skip from yaml is considered as NA, we should update the remediation message to reflect as to why it's being skipped?

pkg/kb-summarizer/summarizer/summarizer.go
```diff
@@ -249,6 +249,9 @@ func (s *Summarizer) processOneResultFileForHost(results *kb.Controls, hostname
...
+			}
 			if msg, ok := s.notApplicable[check.ID]; ok {
```
leodotcloud, line 255: Guess, we can remove `s.notApplicable` map altogether?

pkg/kb-summarizer/summarizer/summarizer.go
```diff
@@ -440,6 +443,9 @@ func (s *Summarizer) loadControls() error {
 				if !check.Scored {
 					continue
 				}
+				if check.Type == CheckTypeSkip {
```
leodotcloud, line 446: Same comments from similar code diff above.

pkg/kb-summarizer/summarizer/summarizer.go
```diff
@@ -486,7 +492,7 @@ func getMappedState(state kb.State) State {
 	case kb.WARN:
 		return Fail
 	case kb.INFO:
-		return Fail
+		return NotApplicable
```
leodotcloud, line 495: Why do we need this change?
Probably we need something like:
```
case NOTAPPLICABLE:
    return NotApplicable
```

leodotcloud (commented):
Are we getting rid of NotApplicable and DefaultSkip information from KDM? If not, why is NA information being moved to yam using skip field?

prachidamle:
> Are we getting rid of NotApplicable and DefaultSkip information from KDM? If not, why is NA information being moved to yam using skip field?

yes @leodotcloud we will no longer have KDM data. So we will be baking in all NotApplicable and DefaultSkip information from KDM into the tests itself using "skip" flag.

leodotcloud:
How are we going to differentiate between default skip and NA with just one `skip: true` flag?

package/Dockerfile
paynejacob: Could we use the source from the release page https://github.com/aquasecurity/kube-bench/archive/v0.3.1.zip

prachidamle:
> How are we going to differentiate between default skip and NA with just one `skip: true` flag?

We are not going to differentiate any more @leodotcloud.

package/Dockerfile
prachidamle: locked down to tag v0.3.1 and with --depth=1

prachidamle: @paynejacob I think with tag and depth we should be good?

paynejacob (approved):

pkg/kb-summarizer/summarizer/summarizer.go
prachidamle: yes, good point, I have appended the reasons from KDM to the remediations

pkg/kb-summarizer/summarizer/summarizer.go
prachidamle: We can, but maybe we need not change the summarizer tool and it can take in the input maps as input - there will not be the defaultSkip and Notapplicable maps provided when Rancher invokes this summarizer tool. But some other cli/client can still run the summarizer by providing these maps. Does that sound ok? @leodotcloud

pkg/kb-summarizer/summarizer/summarizer.go
prachidamle: kb.INFO state is set by kube-bench in the scan output for any tests that are marked as type: "skip"
Thus now that all NA and default skip from KDM are marked as type: "skip", the kube-bench output will mark these tests state as kb.INFO. refer https://github.com/aquasecurity/kube-bench#omitting-checks

Summarizer will need to translate this to NotApplicable so that it shows as NA in the CIS v2 report.

leodotcloud (approved):

paynejacob (approved):

PR #25, opened 2020-09-25
Additional fields to the CIS report

prachidamle:
Adding extra fields needed to generate self-assessment guide.
https://github.com/rancher/cis-operator/issues/37

After this PR is merged, we need to update the cis-operator and rancher-cis-benchmark charts

### DIFF
--- a/pkg/kb-summarizer/report/report.go
+++ b/pkg/kb-summarizer/report/report.go
@@ -3,6 +3,7 @@ package report
 import (
 	"encoding/json"
 	"fmt"
+	"os/exec"
 	"sort"
 
 	"github.com/rancher/security-scan/pkg/kb-summarizer/summarizer"
@@ -28,12 +29,19 @@ const (
 )
 
 type Check struct {
-	ID          string     `yaml:"id" json:"id"`
-	Text        string     `json:"description"`
-	Remediation string     `json:"remediation"`
-	State       State      `json:"state"`
-	NodeType    []NodeType `json:"node_type"`
-	Nodes       []string   `json:"nodes,omitempty"`
+	ID             string      `yaml:"id" json:"id"`
+	Text           string      `json:"description"`
+	Remediation    string      `json:"remediation"`
+	State          State       `json:"state"`
+	NodeType       []NodeType  `json:"node_type"`
+	Nodes          []string    `json:"nodes,omitempty"`
+	Audit          string      `json:"audit"`
+	AuditConfig    string      `json:"audit_config"`
+	TestInfo       []string    `json:"test_info"`
+	Commands       []*exec.Cmd `json:"commands"`
+	ConfigCommands []*exec.Cmd `json:"config_commands"`
+	ActualValue    string      `json:"actual_value"`
+	ExpectedResult string      `json:"expected_result"`
 }
 
 type Group struct {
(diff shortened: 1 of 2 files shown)


pkg/kb-summarizer/report/report.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"sort"
 
+	"os/exec"
+
```
mrajashree, line 9: nit: extra blank lines are preventing imports from being sorted, same in summarizer.go, rest looks good

pkg/kb-summarizer/report/report.go
```diff
@@ -28,12 +30,19 @@ const (
...
+	State          State       `json:"state"`
+	NodeType       []NodeType  `json:"node_type"`
+	Nodes          []string    `json:"nodes,omitempty"`
+	Audit          string      `json:"audit"`
+	AuditConfig    string      `yaml:"audit_config"`
+	TestInfo       []string    `json:"test_info"`
+	Commands       []*exec.Cmd `json:"commands"`
+	ConfigCommands []*exec.Cmd `json:"config_commands"`
+	ActualValue    string      `json:"actual_value"`
+	ExpectedResult string      `json:"expected_result"`
```
aiyengar2, line 45: Is there a reason why the Go tags are `yaml`, `json`, or both for different fields?

If there is a distinction, I think we should move `AuditConfig` to the top since it's the only YAML field.

pkg/kb-summarizer/summarizer/summarizer.go
```diff
@@ -76,16 +78,23 @@ const (
...
+	NodeType       []NodeType                   `json:"t"`
+	NodesMap       map[string]bool              `json:"-"`
+	Nodes          []string                     `json:"n,omitempty"`
+	Audit          string                       `json:"a"`
+	AuditConfig    string                       `yaml:"ac"`
+	TestInfo       []string                     `json:"ti"`
+	Commands       []*exec.Cmd                  `json:"c"`
+	ConfigCommands []*exec.Cmd                  `json:"cc"`
+	ActualValue    string                       `json:"av"`
+	ExpectedResult string                       `json:"er"`
```
aiyengar2, line 97: I'm a little confused about the Go tags here. Is this supposed to be a short form for the tags listed above?

aiyengar2 (approved):
LGTM, Go tags seem a bit weird but I don't think that's part of your changes. Left comments about it either way

pkg/kb-summarizer/summarizer/summarizer.go
prachidamle: Yes I think this is intended to be short-form so that the json file generated by the summarizer is not too huge, this file is anyway intermittent data to be converted by report.go in the rancher report format.
But I will just change them to all json

pkg/kb-summarizer/report/report.go
prachidamle: I changed them to all json, they should not be yaml.

aiyengar2 (dismissed):

pkg/kb-summarizer/report/report.go
prachidamle: Does it look ok now?

pkg/kb-summarizer/summarizer/summarizer.go
```diff
@@ -723,4 +738,12 @@ func printCheckWrapper(cw *CheckWrapper) {
 	logrus.Debugf("node_type: %+v", cw.NodeType)
 	logrus.Debugf("nodes: %+v", cw.Nodes)
 	logrus.Debugf("result: %+v", cw.Result)
+	logrus.Debugf("auditConfig: %+v", cw.AuditConfig)
+	logrus.Debugf("audit: %+v", cw.Audit)
+	logrus.Debugf("commands: %+v", cw.Commands)
+	logrus.Debugf("configcommands: %+v", cw.ConfigCommands)
+	logrus.Debugf("testInfo: %+v", cw.TestInfo)
+	logrus.Debugf("ActualValue: %v", cw.ActualValue)
+	logrus.Debugf("ExpectedResult: %v", cw.ExpectedResult)
```
dramich, line 747: I realize there is a precedent for this pattern but it seems odd to me. Why not use something like [string builder](https://golang.org/pkg/strings/#Builder) to build the debug log and print that? Or just marshall the object and print the result?

prachidamle: I can change this, but actually no where in the cis scan tooling is this code invoked really

prachidamle: I just appended extra fields I am adding - the debug is for anyone who will run this go tool like cli separately.

dramich: Oh, could just do a `%+v ` for the whole object then? No need to change it now

prachidamle: yeah makes sense, changed it to use %+v

pkg/kb-summarizer/report/report.go
aiyengar2: yeah it looks good now

aiyengar2 (approved):

dramich (approved):

PR #27, opened 2020-10-22
Add cis-1.6 profiles

mrajashree:
This PR adds in the upstream CIS-1.6 profile.
https://github.com/rancher/rancher/issues/29649


mrajashree:
Will add more commits to the PR for RKE 1.6 profiles

pkg/kb-summarizer/summarizer/summarizer.go
```diff
@@ -447,9 +449,9 @@ func (s *Summarizer) loadControls() error {
 				s.groupWrappersMap[g.ID] = gw
 			}
 			for _, check := range g.Checks {
-				if !check.Scored {
-					continue
-				}
+				//if !check.Scored {
```
briandowns, line 452: Do we want to leave this commented or just remove it entirely?

mrajashree: this will be removed entirely

mrajashree:
Marking this PR as draft till the changes related to "warn" state are included

mrajashree:
I'm moving the commit to make changes for "warn" and unscored tests to a different PR https://github.com/rancher/security-scan/pull/30

package/cfg/rke-cis-1.6-hardened/config.yaml
```diff
@@ -1,2 +1,10 @@
 ---
 ## Version-specific settings that override the values in cfg/config.yaml
+
+node:
+  kubelet:
+    defaultkubeconfig: "/etc/kubernetes/ssl/kubecfg-kube-node.yaml"
+    defaultcafile: "/etc/kubernetes/ssl/kube-ca.pem"
+
+  proxy:
+    defaultkubeconfig: "/etc/kubernetes/ssl/kubecfg-kube-proxy.yaml"
```
prachidamle, line 10: probably should have a newline?

mrajashree: done

package/cfg/config.yaml
```diff
@@ -243,6 +255,12 @@ target_mapping:
     - "controlplane"
     - "etcd"
     - "policies"
+  "rke-cis-1.6-permissive":
+    - "master"
+    - "node"
+    - "controlplane"
+    - "etcd"
+    - "policies"
```
prachidamle, line 263: We should also add the profiles to the "version_mapping" map in this file: https://github.com/rancher/security-scan/blob/master/package/cfg/config.yaml#L191

mrajashree: added it in, changed default profile to CIS 1.6 for k8s 1.16 onwards as per discussion offline with you and @cloudnautique

prachidamle: Just putting a note here: we will have to add new scanprofile and benchmark CRs in the chart for cis-1.6 profiles.

package/cfg/config.yaml
```diff
@@ -191,14 +191,11 @@ managedservices:
...
-  "rke-1.18": "rke-cis-1.5-hardened"
   "rke-1.15": "rke-cis-1.5-permissive"
-  "rke-1.16": "rke-cis-1.5-permissive"
-  "rke-1.17": "rke-cis-1.5-permissive"
-  "rke-1.18": "rke-cis-1.5-permissive"
+  "rke-1.16": "rke-cis-1.6-permissive"
+  "rke-1.17": "rke-cis-1.6-permissive"
+  "rke-1.18": "rke-cis-1.6-permissive"
+  "rke-1.19": "rke-cis-1.6-permissive"
   "v1.18.10+rke2r1": "rke2-cis-1.5-hardened"
```
prachidamle, line 199: @briandowns While reviewing this realised that we need to set only one profile as the default for each k8s version. Thus this should change for rke2 to not have both permissive and hardened mapped to "v1.18.10+rke2r1"

prachidamle (approved):
LGTM

package/cfg/config.yaml
briandowns: Thank you for the heads up. I'll update this and test in my next PR unless you're prefer it in a separate PR that addresses only this.

PR #30, opened 2020-11-16
Add "warn" state, include unscored tests in report

mrajashree:
This commit changes the report generation logic to include all tests, scored and unscored.
It also adds a new state "warn" corresponding to the "warn" output that kube-bench generates
for some of the manual tests.
It adds two new columns to the report, one to indicate if the test is scored/unscored and another
one to show if test is manual/automated

https://github.com/rancher/cis-operator/issues/45
https://github.com/rancher/cis-operator/issues/42

### DIFF
--- a/pkg/kb-summarizer/summarizer/summarizer.go
+++ b/pkg/kb-summarizer/summarizer/summarizer.go
@@ -59,6 +59,7 @@ const (
 	Skip          State = "S"
 	Mixed         State = "M"
 	NotApplicable State = "N"
+	Warn          State = "W"
 
 	SKIP kb.State = "SKIP"
 	NA   kb.State = "NA"
@@ -78,10 +79,10 @@ const (
 type CheckWrapper struct {
 	ID             string                       `yaml:"id" json:"id"`
 	Text           string                       `json:"d"`
-	Type           string                       `json:"-"`
+	Type           string                       `json:"tt"`
 	Remediation    string                       `json:"r"`
 	State          State                        `json:"s"`
-	Scored         bool                         `json:"-"`
+	Scored         bool                         `json:"sc"`
 	Result         map[kb.State]map[string]bool `json:"-"`
 	NodeType       []NodeType                   `json:"t"`
 	NodesMap       map[string]bool              `json:"-"`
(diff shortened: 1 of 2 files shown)


briandowns (approved):
LGTM

pkg/kb-summarizer/summarizer/summarizer.go
```diff
@@ -78,10 +79,10 @@ const (
 type CheckWrapper struct {
 	ID             string                       `yaml:"id" json:"id"`
 	Text           string                       `json:"d"`
-	Type           string                       `json:"-"`
+	Type           string                       `json:"tt"`
```
prachidamle, line 82: Where do we set the type?

pkg/kb-summarizer/summarizer/summarizer.go
```diff
@@ -78,10 +79,10 @@ const (
...
 	Remediation    string                       `json:"r"`
 	State          State                        `json:"s"`
-	Scored         bool                         `json:"-"`
+	Scored         bool                         `json:"sc"`
```
prachidamle, line 85: Do we need to set this?

pkg/kb-summarizer/summarizer/summarizer.go
mrajashree: at this line https://github.com/rancher/security-scan/pull/30/files#diff-cdbf6a6affc59d13e5fa37a2e5a29596d90e108f7bf5cc20a29e7d0929d56069R121

pkg/kb-summarizer/summarizer/summarizer.go
mrajashree: hm the first time I tried it I had made this change, I can try without it

prachidamle: nevermind, I meant where are we assigning a value to this - I see it is done in https://github.com/rancher/security-scan/blob/a1771f4ac81091e73723168014ae0cd93f04021d/pkg/kb-summarizer/summarizer/summarizer.go#L504

prachidamle (approved):
LGTM

Issue #29, opened 2020-11-13
Audit field in ClusterScanReport .spec.reportJSON not showing interpolated variables


paraglade:
The interpolated values of variables in the the audit field are not being rendered in the ClusterScanReport `.spec.reportJSON` for example:
`"audit": "/bin/sh -c 'if test -e $apiserverconf; then stat -c %a $apiserverconf; fi'",`

Should be:
`"audit": "/bin/sh -c 'if test -e /etc/kubernetes/manifests/kube-apiserver.yaml; then stat -c permissions=%a /etc/kubernetes/manifests/kube-apiserver.yaml; fi'"`

in the `reportJSON`

prachidamle:
@paraglade These variables will have different values depending on the cluster types (rke1, rke2, k3s) and is cluster setup dependent (with custom benchmarks it might be something different than what we know of )- how will we know which value to use to replace the $apiserverconf

prachidamle:
@maggieliu maybe this should move to 2.5.4 since it is being fixed with https://github.com/rancher/security-scan/issues/28

paraglade:
confirmed interpolated data is in the report json and I am able to generate the needed docs.
tested with:`rancher-cis-benchmark:1.0.300`

Issue #28, opened 2020-11-13
Missing data in ClusterScanReport .spec.reportJSON


paraglade:
The following fields in the ClusterScanReport located at `.spec.reportJSON.results[].checks[]` are missing data
```
audit_config
actual_value
expected_result
```
This data is needed to produce the self-assessment documentation.

prachidamle:
This data is read from the Kube-bench reports itself - so it seems the KB report does not print out this data. We might need to check the KB cli to see if there is a flag to output it.

paraglade:
confirmed required data is in the report json and I am able to generate the needed docs.
tested with:`rancher-cis-benchmark:1.0.300`

PR #33, opened 2020-12-08
Copy over the actual result info of the test after running the scan from the kube-bench results

prachidamle:
The information from the kube-bench results was not getting copied while creating the scan report - instead the test config as found in the config yamls was left as is after being initialized in loadControls(). Hence the data of fields like these was missing :
- audit_config
- actual_value
- expected_result

And also the interpolated values of variables in the the audit field were not being rendered in the final ClusterScanReport.

https://github.com/rancher/security-scan/issues/28
https://github.com/rancher/security-scan/issues/29

### DIFF
--- a/pkg/kb-summarizer/summarizer/summarizer.go
+++ b/pkg/kb-summarizer/summarizer/summarizer.go
@@ -77,23 +77,23 @@ const (
 )
 
 type CheckWrapper struct {
-	ID             string                       `yaml:"id" json:"id"`
-	Text           string                       `json:"d"`
-	Type           string                       `json:"tt"`
-	Remediation    string                       `json:"r"`
-	State          State                        `json:"s"`
-	Scored         bool                         `json:"sc"`
-	Result         map[kb.State]map[string]bool `json:"-"`
-	NodeType       []NodeType                   `json:"t"`
-	NodesMap       map[string]bool              `json:"-"`
-	Nodes          []string                     `json:"n,omitempty"`
-	Audit          string                       `json:"a"`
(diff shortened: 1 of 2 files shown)


mrajashree (approved):
lgtm!

pkg/kb-summarizer/summarizer/summarizer.go
```diff
@@ -638,6 +641,19 @@ func (s *Summarizer) runFinalPassOnCheckWrapper(cw *CheckWrapper) {
...
 }
 
+func (s *Summarizer) copyDataFromResults(cw *CheckWrapper) {
+	checkFromResults := s.checkWrappersMaps[cw.ID]
+	if checkFromResults == nil {
+		return
+	}
+	cw.Audit = checkFromResults.Audit
+	cw.AuditConfig = checkFromResults.AuditConfig
+	cw.ActualValue = checkFromResults.ActualValue
```
leodotcloud, line 651: If the same check runs on two nodes with a pass on one and a fail on the other one, actualValue would be overwritten, right?

prachidamle: oh, so then should be a map <nodeIP>:<value>?

leodotcloud: For interpolation of the variables, it's ok to copy over from the result. But the "actual value" is per node, so we might have to store using a map.

prachidamle: Added a map

mrajashree (approved):

dramich (approved):

PR #34, opened 2020-12-09
Changes to some tests audit config

mrajashree:
This PR adds following changes:
* The tests 1.2.10, 5.2.2, 5.2.3, 5.2.4, 5.2.5, 5.3.2 and 5.7.4 have been automated in hardened profiles
and pass on hardened clusters. But they don't pass on non-hardened clusters and generate "Warn" output.
This commit skips them on the permissive profile so they don't generate "Warn" output. They were skipped for the 1.5 profile, doing the same for 1.6 profile too

https://github.com/rancher/rancher/issues/29649#issuecomment-733985659

* For test 1.1.12 the RKE hardening guide provides steps on configuring etcd data directory ownership. So this test should be skipped on a non-hardened cluster since the ownership won't be configured on it. Also this test along with tests 1.1.11, 1.1.19, 1.1.20 and 1.1.21 was originally in master.yaml in the upstream profiles. But for RKE 1.5 profiles these 5 tests were moved from master.yaml to etcd.yaml to be run on the etcd nodes. This PR does the same for the RKE 1.6 profile moving these tests from master.yaml to etcd.yaml
https://github.com/rancher/cis-operator/issues/62

* Test 4.2.8 is to be skipped on all profiles, because RKE sets --hostname-override. This PR also adds the reason in remediation

### DIFF
--- a/package/cfg/rke-cis-1.5-hardened/node.yaml
+++ b/package/cfg/rke-cis-1.5-hardened/node.yaml
@@ -502,6 +502,7 @@ groups:
         # This is one of those properties that can only be set as a command line argument.
         # To check if the property is set as expected, we need to parse the kubelet command
         # instead reading the Kubelet Configuration file.
+        type: "skip"
         audit: "/bin/ps -fC $kubeletbin "
         tests:
           test_items:
@@ -514,6 +515,8 @@ groups:
           Based on your system, restart the kubelet service. For example:
           systemctl daemon-reload
           systemctl restart kubelet.service
+
+          Clusters provisioned by RKE set the --hostname-override to avoid any hostname configuration errors
         scored: false
 
       - id: 4.2.9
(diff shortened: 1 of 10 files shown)


package/cfg/rke-cis-1.5-hardened/node.yaml
```diff
@@ -502,6 +502,7 @@ groups:
         # This is one of those properties that can only be set as a command line argument.
         # To check if the property is set as expected, we need to parse the kubelet command
         # instead reading the Kubelet Configuration file.
+        type: "skip"
```
prachidamle, line 505: We always had this in 1.5 profiles, but unscored, do we need to do this now because we are including scored/unscored both tests?

mrajashree: @prachidamle we added 4.2.8 as a part of adding Unscored tests to 1.5 profiles with this https://github.com/rancher/security-scan/commit/65db75822820e05106bb6d6f0a2763b69c7e3122
And this test generates warn output because RKE uses --hostname-override, so I thought we could skip it

prachidamle: ok, I think we can skip it since I found some discussion around skipping it.

prachidamle (approved):
LGTM

dramich (approved):

PR #35, opened 2020-12-11
Set type to manual for all manual tests

mrajashree:
Most of the manual tests, that have the word (Manual) in the text field are marked
manual by setting type to "manual". Some of these tests don't have type field set to anything.
The new column in cis report stays empty for such tests. This commit ensures any test which has
the word "Manual" in its text field is marked type:manual, so that all tests with an empty
test_type column can be assumed to be automated.

https://github.com/rancher/cis-operator/issues/45#issuecomment-742046103


prachidamle (approved):
LGTM

PR #42, opened 2021-04-22
Add k3s-cis permissive and hardened profiles 

MonzElmasry:
issue :  k3s-io/k3s#1504

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -17,7 +17,10 @@ RUN apt-get update \
     && apt-get install -y \
     curl \
     jq \
-    vim
+    vim \
+    systemd \
+    grep \
+    bash
 RUN curl -sLf https://storage.googleapis.com/kubernetes-release/release/v1.13.5/bin/linux/amd64/kubectl > /usr/local/bin/kubectl-1.13 \
     && ln -sv /usr/local/bin/kubectl-1.13 /usr/local/bin/kubectl \
     && chmod +x /usr/local/bin/kubectl*
--- a/package/cfg/config.yaml
+++ b/package/cfg/config.yaml
@@ -25,6 +25,7 @@ master:
       - "hyperkube apiserver"
       - "hyperkube kube-apiserver"
       - "apiserver"
+      - "containerd"
     confs:
       - /etc/kubernetes/manifests/kube-apiserver.yaml
       - /etc/kubernetes/manifests/kube-apiserver.manifest
@@ -38,6 +39,7 @@ master:
       - "hyperkube scheduler"
       - "hyperkube kube-scheduler"
       - "scheduler"
+      - "containerd"
     confs:
       - /etc/kubernetes/manifests/kube-scheduler.yaml
       - /etc/kubernetes/manifests/kube-scheduler.manifest
@@ -52,6 +54,7 @@ master:
       - "hyperkube controller-manager"
       - "hyperkube kube-controller-manager"
       - "controller-manager"
+      - "containerd"
     confs:
       - /etc/kubernetes/manifests/kube-controller-manager.yaml
       - /etc/kubernetes/manifests/kube-controller-manager.manifest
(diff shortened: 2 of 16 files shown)


galal-hussein (approved):

package/Dockerfile
```diff
@@ -17,7 +17,10 @@ RUN apt-get update \
     && apt-get install -y \
     curl \
     jq \
-    vim
+    vim \
+    systemd \
```
brandond, line 21: why are we installing systemd in a Docker image?

galal-hussein: to basically get the journalctl logs from k3s, to get related flags to different components, since there are no logs or argument file for those

package/helper_scripts/check_for_k3s_network_policies.sh
```diff
@@ -0,0 +1,19 @@
...
+    echo "false"
+}
+
+trap 'handle_error' ERR
+
+for namespace in kube-system kube-public default; do
+  policy_count=$(/var/lib/rancher/k3s/bin/kubectl get networkpolicy -n ${namespace} -o json | jq -r '.items | length')
+  if [ ${policy_count} -eq 0 ]; then
+    echo "false"
+    exit
```
nickgerace, line 15: Should we include an exit code here?

package/cfg/k3s-cis-1.6-hardened/config.yaml
```diff
@@ -0,0 +1,46 @@
...
+  components:
+    - apiserver
+    - scheduler
+    - controllermanager
+    - etcd
+    - policies
+
+  apiserver:
+    bins:
+      - containerd
```
galal-hussein, line 14: clarification on using containerd binary: there is no way to use either k3s or k3s-server in the binary here, since the cmd for the k3s process is `k3s-server` (/proc/<k3s-pid>/cmd) and kubebench check for this cmd and compare it with the process name which is `/usr/local/bin/k3s server` so the safest bet is to use containred instead

package/Dockerfile
brandond: Ah, so we are installing all of systemd just to get journalctl? Is there a minimal package available for that?

package/cfg/config.yaml
```diff
@@ -25,6 +25,7 @@ master:
       - "hyperkube apiserver"
       - "hyperkube kube-apiserver"
       - "apiserver"
+      - "containerd"
```
prachidamle, line 28: Is it possible that there might containerd process found on a node which is not a k3s node? is there any other k3s specific process or path that can be used?

package/run_sonobuoy_plugin.sh
```diff
@@ -52,7 +56,7 @@ mkdir -p "${RESULTS_DIR}"
 
 # etcd
 if [[ "${OVERRIDE_BENCHMARK_VERSION}" != "" ]]; then
-  if [[ "$(pgrep -f /etcd | wc -l)" -gt 0 ]]; then
+  if [[ "$(pgrep -f /etcd | wc -l)" -gt 0 ]]  || [[ "$(journalctl -D /var/log/journal -u k3s | grep 'Managed etcd' | grep -v grep | wc -l)" -gt 0 ]]; then
```
prachidamle, line 59: Can the log path be other than /var/log/journal? Maybe we can pass this as an env var?

brandond (dismissed):

briandowns (approved):

galal-hussein (approved):

package/cfg/config.yaml
MonzElmasry: we tried binaries other than containerd and it doesnt work in k3s environment

package/run_sonobuoy_plugin.sh
```diff
@@ -45,7 +44,7 @@ RANCHER_K8S_VERSION="${K8S_API_VERSION}"
 echo "Rancher Kubernetes Version: ${RANCHER_K8S_VERSION}"
 
 set -x
-
+JOURNAL_LOG="/var/log/journal"
```
prachidamle, line 47: Shouldn't this be in the format `CONFIG_DIR="${CONFIG_DIR:-/etc/kube-bench/cfg}"` - so that read from env if present else use the default?

prachidamle (approved):
LGTM

briandowns (approved):

PR #50, opened 2021-07-28
Fix to K3s checks on non-etcd cluster

dereknola:
Associated Issue: https://github.com/rancher/cis-operator/issues/100

This PR removes the if check around the etcd.yaml benchmarks, this was causing MIXED results on the scan. Now, all etcd related checks are run thru a helper script, which if etcd is running, performs the proper check. If etcd is not running, like on a single node using sqlite3, then the benchmarks are still run, but the helper script returns the correct answer for the check to pass. 

Both k3s permissive and hardened cis-1.6 scans now pass, on both etcd and non-etcd clusters.
Screenshot from 2021-07-28 13-33-53
Screenshot from 2021-07-27 11-32-20
Screenshot from 2021-07-27 11-32-31
Screenshot from 2021-07-28 12-25-46

### DIFF
--- a/package/run_sonobuoy_plugin.sh
+++ b/package/run_sonobuoy_plugin.sh
@@ -59,7 +59,7 @@ mkdir -p "${RESULTS_DIR}"
 
 # etcd
 if [[ "${OVERRIDE_BENCHMARK_VERSION}" != "" ]]; then
-  if [[ "$(pgrep -f /etcd | wc -l)" -gt 0 ]]  || [[ "$(journalctl -D $JOURNAL_LOG -u k3s | grep 'Managed etcd' | grep -v grep | wc -l)" -gt 0 ]]; then
+  if [[ "$(pgrep -f /etcd | wc -l)" -gt 0 ]]  || [[ "$(journalctl -D $JOURNAL_LOG -u k3s | wc -l)" -gt 0 ]]; then
     echo "etcd: Using OVERRIDE_BENCHMARK_VERSION=${OVERRIDE_BENCHMARK_VERSION}"
     kube-bench run \
       --targets etcd \
--- a/README.md
+++ b/README.md
@@ -11,8 +11,25 @@ Multi-purpose repo:
 The corresponding docker image (rancher/security-scan) is used in the system charts.
 
 ## Building
-
 `make`
+
+Tag the image to personal docker hub repo
+
+`docker tag rancher/security-scan:<MAKE TAG OUTPUT> <DOCKER_HUB_USER>/security-scan:dev`
+
+Push docker tag
+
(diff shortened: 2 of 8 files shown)


briandowns (approved):

package/run_sonobuoy_plugin.sh
```diff
@@ -59,7 +59,6 @@ mkdir -p "${RESULTS_DIR}"
 
 # etcd
 if [[ "${OVERRIDE_BENCHMARK_VERSION}" != "" ]]; then
-  if [[ "$(pgrep -f /etcd | wc -l)" -gt 0 ]]  || [[ "$(journalctl -D $JOURNAL_LOG -u k3s | grep 'Managed etcd' | grep -v grep | wc -l)" -gt 0 ]]; then
```
prachidamle, line 62: This impacts all rke1/rke2/k3s or any custom benchmark scan. This check makes sure for rke1 that etcd benchmarks are run only on nodes having etcd running. So I dont think we should be deleting this out.

package/run_sonobuoy_plugin.sh
```diff
@@ -72,9 +71,7 @@ if [[ "${OVERRIDE_BENCHMARK_VERSION}" != "" ]]; then
       --json \
       --log_dir "${LOG_DIR}" \
       --outputfile "${RESULTS_DIR}/etcd.json" 2> "${ERROR_LOG_FILE}"
-  fi
 else
-  if [[ "$(pgrep -f /etcd | wc -l)" -gt 0 ]]; then
     echo "etcd: Using RANCHER_K8S_VERSION=${RANCHER_K8S_VERSION}"
```
prachidamle, line 75: Same comment as above. This will break in multi-node clusters where etcd is running only one few and we want to run etcd benchmarks only on the nodes running etd

prachidamle: We need to test this with rke1 (multi-node with varied roles), rke2, gke/eks and k3s. All scan profiles need to pass before we change this common area.

dereknola: This has been updated always run when k3s is running. This should limit the impact to other distributions.

package/run_sonobuoy_plugin.sh
dereknola: This has been updated to only affect the k3s side of this check. If k3s is running we now always pass. This should limit the impact to other distributions.

prachidamle:
Changes look good, please can you squash all commits and push again?

dereknola:
@prachidamle Squashed.

prachidamle (approved):
LGTM

briandowns (approved):
LGTM

PR #58, opened 2021-10-11
[2.6] fixed k3s check for etcd

paynejacob:
https://github.com/rancher/cis-operator/issues/108

# Problem
The logic that checks for k3s clusters was not looking at the agent logs.  This caused scans to fail due to missing results.

# Solution
Added a check for agent logs to the etcd check

# Testing

- tested on an HA k3s cluster with a sql backend
- tested on HA k3s with embedded etcd
- tested on single node k3s

all 3 clusters reported scan failures with the current image.  With these changes in a new rancher-security image all tests passed.


dereknola (approved):

rmweir (commented):
LGTM

prachidamle:
@paynejacob can you please describe what error was seen in the Sonobuoy run? Were we skipping etcd checks on nodes running k3s agent?

PR #64, opened 2021-11-24
Update node.yaml to fix some issues regarding file permissions on 4.1.5 and 4.1.7 checks

frjaraur:
Issues with file permissions on 4.1.5 (we can use more restrictive) and 4.1.7 (error on check vs message)


andypitcher:
Closing this pull request, for these two reasons:

1. Even if this change is on point, CIS Workbench recommended 644, instead of 600, see [4.1.5](https://github.com/aquasecurity/kube-bench/blob/c3e3c4c31c26e6dddfe937e46f3f0dc8c98aa774/cfg/cis-1.5/node.yaml#L70) and [4.1.7](https://github.com/aquasecurity/kube-bench/blob/c3e3c4c31c26e6dddfe937e46f3f0dc8c98aa774/cfg/cis-1.5/node.yaml#L102) in kube-bench. The permissions have been set to 600 starting CIS-1.24.
2. CIS-1.5 covers k8s 1.15 which is EOL

PR #71, opened 2022-02-23
Add set flag for rke-cis-1.6-hardened profile

dhruvmewada15:
This PR  adds `set` flag in `rke-cis-1.6-hardened` profile for 1.3.7 and 1.4.2 tests which was missing.


prachidamle (approved):
LGTM!

PR #72, opened 2022-03-31
enable fossa scanning

paraglade:
request: https://github.com/rancherlabs/eio/issues/772


samkulkarni20 (changes requested):
The drone builds are failing. Will wait till they're fixed.

PR #76, opened 2022-05-05
Add cis-1.20 profiles for rke1

dhruvmewada15:
This PR introduces upstream [cis-1.20](https://github.com/aquasecurity/kube-bench/tree/main/cfg/cis-1.20) benchmark profiles for the security scan for rke1 cluster :

- `rke-cis-1.20-permissive`
- `rke-cis-1.20-hardened`

Linked Issue:
https://github.com/rancher/cis-operator/issues/135

### DIFF
--- a/package/cfg/config.yaml
+++ b/package/cfg/config.yaml
@@ -207,6 +207,8 @@ version_mapping:
   "rke-1.16": "rke-cis-1.6-permissive"
   "rke-1.17": "rke-cis-1.6-permissive"
   "rke-1.18": "rke-cis-1.6-permissive"
+  "rke-1.19": "rke-cis-1.20-permissive"
+  "rke-1.20": "rke-cis-1.20-permissive"
   "v1.18.10+rke2r1": "rke2-cis-1.5-hardened"
   "v1.18.10+rke2r1": "rke2-cis-1.5-permissive"
   "eks-1.0": "eks-1.0"
@@ -280,6 +282,12 @@ target_mapping:
     - "controlplane"
     - "etcd"
     - "policies"
+  "rke-cis-1.20-permissive":
+    - "master"
+    - "node"
+    - "controlplane"
+    - "etcd"
+    - "policies"  
   "eks-1.0":
     - "node"
   "gke-1.0":
--- a/package/cfg/config.yaml
+++ b/package/cfg/config.yaml
@@ -264,6 +264,12 @@ target_mapping:
     - "controlplane"
     - "etcd"
     - "policies"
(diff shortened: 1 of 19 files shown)


dhruvmewada15:
RKE1 single node non-hardened cluster:
image

RKE1 single node hardened cluster:
image

RKE1 multi node non-hardened cluster:
image


RKE1 multi node hardened cluster:
image

package/cfg/config.yaml
```diff
@@ -207,7 +207,8 @@ version_mapping:
   "rke-1.16": "rke-cis-1.6-permissive"
   "rke-1.17": "rke-cis-1.6-permissive"
   "rke-1.18": "rke-cis-1.6-permissive"
-  "rke-1.19": "rke-cis-1.6-permissive"
+  "rke-1.19": "rke-cis-1.20-permissive"
+  "rke-1.20": "rke-cis-1.20-permissive"
```
macedogm, line 211: Quick question, shouldn't we also add the hardened profile `rke-cis-1.20-hardened` here?

dhruvmewada15: kube-bench requires either `kubernetes version` OR `benchmark version` when running the test. In our case we run CIS scan through UI only and there we are always passing the benchmark version/profiles (For example `rke-cis-1.6-hardened`) so we dont need to actually update the version_mapping block. @prachidamle Is my understanding correct?

prachidamle: This version_mapping block specifies the default benchmark version to be used for a given kubernetes version for the cluster type. We should update it to point to the default profile to be used in scanning, which will be the permissive profile.

prachidamle:
This looks good, one quick question - were there any new tests added in CIS 1.20 as compared to CIS 1.16?

dhruvmewada15:
> This looks good, one quick question - were there any new tests added in CIS 1.20 as compared to CIS 1.16?

There are few tests added in _policies_ and also tests removed in _master_. Also at many places it has corrections for remediations or test desciptions.

prachidamle:
Also we will need a charts PR to add these clusterscanprofiles and clusterscanbenchmark CRs for CIS 1.20? https://github.com/rancher/charts/tree/dev-v2.6/packages/rancher-cis-benchmark/charts/templates

prachidamle (approved):
lgtm, we need a charts PR though.

macedogm:
LGTM

I tested both `RKE single node non-hardened cluster` and `RKE single node hardened cluster` and the scans passed as expected.

I haven't tested the `multi node` setup.

PR #85, opened 2022-06-16
Update to kube-bench upstream.

rayandas:
This PR updates kube-bench to `v0.6.8`, sonobuoy to `0.56.7` and Go version to `1.17` (and other necessary changes because of the version upgrade.) in Security scan. 
Part of: https://github.com/rancher/cis-operator/issues/142
Built the image with latest changes and used that image to scan `rke-profile-permissive-1.6` profile. And it's working fine. See the attachment. 
Screenshot from 2022-06-16 13-41-26
There are few failures which are because of the version bump (Go `1.12` to `1.17` and kube-bench `0.2.3-rancher-3` to `0.6.8`), we can raise another issue to fix those failures.


dhruvmewada15 (approved):
LGTM

prachidamle:
LGTM, but what are the failures in the permissive profile? We should also fix the permissive profile failures before we can release this change.

rayandas:
@prachidamle There are failures related to permissions. We're working on fixing the failures right now (RKE1, RKE2, and K3S). We created https://github.com/rancher/cis-operator/issues/156 to track it.

scripts/validate
```diff
@@ -16,5 +16,5 @@ for i in ${PACKAGES}; do
     fi
 done
 test -z "$failed"
-echo Running: go fmt
-test -z "$(go fmt ${PACKAGES} | tee /dev/stderr)"
+#echo Running: go fmt
+#test -z "$(go fmt ${PACKAGES} | tee /dev/stderr)"
```
macedogm, line 20: Is this comment on purpose or just some leftover from tests?

pkg/kb-summarizer/summarizer/summarizer.go
```diff
@@ -600,6 +600,7 @@ func (s *Summarizer) runFinalPassOnCheckWrapper(cw *CheckWrapper) {
 	nodesMap := s.getNodesMapOfCheckWrapper(cw)
 	nodeCount := len(nodesMap)
 	logrus.Debugf("id: %s nodeCount: %d", cw.ID, nodeCount)
+	fmt.Println(cw.Result)
```
macedogm, line 603: Is this print statement intended or only for testing?

scripts/validate
rayandas: ah! it was some leftover from test. Updating the PR.

pkg/kb-summarizer/summarizer/summarizer.go
rayandas: This was for testing. updating the PR after removing this.

prachidamle (dismissed):

rayandas:
Rebased with upstream. That fixes the CI and update my PR. @prachidamle Can I get another approval? Then we can merge this.

samkulkarni20 (approved):
Approving the rebase. PR changes were previously approved by @prachidamle

rayandas:
merging this PR.

leodotcloud:
@rayandas Just happened to stumble upon this PR while looking at something else. Was `__debug_bin` file under `cmd/kb-summarizer` checked in accidentally?

rayandas:
Hi @leodotcloud I think so, yes. It should be in `.gitignore` to avoid pushing such files. Should I raise a PR to remove it and add it in .gitignore?

leodotcloud:
Yes, that would be helpful @rayandas!

PR #86, opened 2022-06-16
Fix for failing tests for rke1 scan profiles in multi node clusters

dhruvmewada15:
This PR fixes tests failing in multi node cluster for profiles:

- rke-profile-permissive-1.5
- rke-profile-permissive-1.6
- rke-profile-hardened-1.5
- rke-profile-hardened-1.6

This PR updates the `if` condition in `run_sonobuoy_plugin.sh` so that it can identify etcd node and be able to run etcd related checks on the node. Earlier condition was causing intermittent results due to which sometimes scan was failing because etcd checks could not be run on the etcd nodes. Have also tested this on RKE2 clusters and its working fine there as well. 

Linked Issue:
https://github.com/rancher/rancher/issues/37944

### DIFF
--- a/package/run_sonobuoy_plugin.sh
+++ b/package/run_sonobuoy_plugin.sh
@@ -57,14 +57,9 @@ if [[ "$(journalctl -D $JOURNAL_LOG --lines=0 2>&1 | grep -s 'No such file or di
 fi
 mkdir -p "${RESULTS_DIR}"
 
-ETCD_PROC="etcd"
-if [ ! ${IS_RKE2} ]; then
-  ETCD_PROC="/etcd"
-fi
-
 # etcd
 if [[ "${OVERRIDE_BENCHMARK_VERSION}" != "" ]]; then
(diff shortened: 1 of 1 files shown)


dhruvmewada15:
RKE1 multi node downstream hardened cluster:
image

rancher-max (commented):
I'm not as familiar with rke1, but my only concern is that this doesn't break RKE2 since it's defined slightly differently above:
```
ETCD_PROC="etcd"
if [ ! ${IS_RKE2} ]; then
  ETCD_PROC="/etcd"
fi
```

Are you able to run an rke2 scan as well just to confirm?

dhruvmewada15:
> I'm not as familiar with rke1, but my only concern is that this doesn't break RKE2 since it's defined slightly differently above:
> 
> ```
> ETCD_PROC="etcd"
> if [ ! ${IS_RKE2} ]; then
>   ETCD_PROC="/etcd"
> fi
> ```
> 
> Are you able to run an rke2 scan as well just to confirm?

@rancher-max Yes we have checked in RKE2 both multi node and single node cluster configuration and it is working as expected. Thanks for pointing out this block to check if its rke2 cluster. We have removed it as its not needed anymore.

rancher-max (approved):
LGTM

PR #87, opened 2022-06-29
Fix security-scan pod log errors.

doflamingo721:
Linked issue:
https://github.com/rancher/cis-operator/issues/149

### DIFF
--- a/package/run.sh
+++ b/package/run.sh
@@ -28,7 +28,7 @@ handle_error() {
       sleep infinity
   fi
   # Annotate self (pod) to signal "error"
-  if [[ -f "${ERROR_LOG_FILE}" ]]; then
+  if [[ -s "${ERROR_LOG_FILE}" ]]; then
       if ! kubectl -n "${SONOBUOY_NS}" \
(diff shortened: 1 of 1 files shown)


package/run.sh
```diff
@@ -28,7 +28,7 @@ handle_error() {
       sleep infinity
   fi
   # Annotate self (pod) to signal "error"
-  if [[ -f "${ERROR_LOG_FILE}" ]]; then
+  if [[ -s "${ERROR_LOG_FILE}" ]]; then
```
prachidamle, line 31: Can you explain why this change is needed?

doflamingo721: `-f` checks if the file is a regular file or not and returns true if the file is a regular file (not a directory). So even if the error log file is empty, it will still return true because the file exists. Where as `-s` checks if file is not zero size. So if the error log file is empty, it won't execute the block.

prachidamle (approved):
Thanks for debugging this one, LGTM

dhruvmewada15 (approved):
LGTM

PR #92, opened 2022-07-22
Fix v1.24 Cluster Scan Failures

rayandas:
After QA validation, they found that the --insecure-port check is failing on every cluster. So this PR fixes them all.
check_cafile_permissions.sh script is also failing (on K8s 1.24 cluster) for RKE1 permissive profile scans, this PR fixes that as well.
Related:
https://github.com/rancher/cis-operator/issues/135
https://github.com/rancher/cis-operator/issues/153
Note: All the fixes are tested on respective clusters.

### DIFF
--- a/package/cfg/k3s-cis-1.20-permissive/master.yaml
+++ b/package/cfg/k3s-cis-1.20-permissive/master.yaml
@@ -615,13 +615,16 @@ groups:
 
       - id: 1.2.18
         text: "Ensure that the --insecure-port argument is set to 0 (Automated)"
-        audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1 | grep 'insecure-port'"
+        audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1"
         tests:
+          bin_op: or
           test_items:
             - flag: "--insecure-port"
               compare:
                 op: eq
                 value: 0
+            - flag: "--insecure-port"
+              set: false
         remediation: |
           Edit the API server pod specification file $apiserverconf
           on the master node and set the below parameter.
--- a/package/cfg/k3s-cis-1.6-permissive/master.yaml
+++ b/package/cfg/k3s-cis-1.6-permissive/master.yaml
@@ -595,13 +595,16 @@ groups:
 
       - id: 1.2.19
         text: "Ensure that the --insecure-port argument is set to 0 (Automated)"
-        audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1 | grep 'insecure-port'"
+        audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1"
         tests:
+          bin_op: or
           test_items:
             - flag: "--insecure-port"
               compare:
                 op: eq
                 value: 0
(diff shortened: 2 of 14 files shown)


macedogm (approved):

rayandas:
merging the PR.

samkulkarni20 (approved):

rancher-max:
I don't believe that setting an or condition to check "whether the `insecure-port` flag is either set to 0 or not set at all" is sufficient. The flag is removed as part of 1.24+ (see https://github.com/rancher/cis-operator/issues/158), but is still present for all previous releases and has a default value of 8080.

The change in this PR could therefore result in falsely passing scans (both permissive and hardened) for any minor release prior to 1.24.

mitulshah-suse:
Referring to the https://github.com/kubernetes/kubernetes/issues/91506#issuecomment-755750903 it looks like this flag was removed in 1.21 itself and only allows it to be set to 0 if at all it is used. 

This flag was deprecated back in 1.10 https://github.com/kubernetes/kubernetes/issues/91506#issue-625971337

Also, none of the kubernetes documentation from 1.20+ has this flag as an option for apiserver.
https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/
https://v1-23.docs.kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/
https://v1-22.docs.kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/
https://v1-21.docs.kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/
https://v1-20.docs.kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/

So, I think we are fine for versions 1.21+
Are we expecting different scan conditions for different kubernetes versions? On clusters below 1.21 the flag should mandatorily be set to 0 and on versions above the or condition currently defined will suffice?
@rancher-max Please advise

rancher-max:
You're right that the flag itself was semi-removed in 1.21. My previous message was therefore somewhat incorrect. The problem then is similar though: "The change in this PR could therefore result in falsely passing scans for any minor release prior to **1.21**.

From what I understand, our profiles conform to the [CIS Benchmark](https://downloads.cisecurity.org/) guides. The starred one in the below screenshot is our default:
image


I don't think it's correct to necessarily change what our checks are doing to be different from these. Rather we should properly to document to users that "if they are using kubernetes v1.23+, they should use the 1.23 scan profile". I obtained the following information from these PDFs:

- ## The 1.20 guide DOES include a check for the insecure-port arg:
image


- ## The 1.23 guide does **NOT** include a check for the insecure-port arg

- ## The default 1.6 guide DOES include a check for the insecure-port arg:
image

---

### To me, it's almost more correct if we **don't** have this OR condition that was added as part of the PR, and instead actually **keep the passing scans**.

mitulshah-suse:
So please confirm if we should do the below changes

- For 1.20 and 1.6, we have a mandatory check for insecure port flag without the OR condition
- For 1.23 we remove the scan itself, since it it not part of the guide.

Please note that this will show **Failed** scans for 1.20 and 1.6 on clusters where this flag is missing. So please confirm that this is the expected result and we are fine with those failures.

rancher-max:
I think that's the correct approach, but I am not the right person to make the call. Adding @prachidamle @brandonsuse @samkulkarni20 @MKlimuszka for visibility/prioritization.

PR #94, opened 2022-09-08
move psp hardened rke2 tests to psa

galal-hussein:


dereknola (approved):

brandond (approved):

briandowns (approved):

package/cfg/rke2-cis-1.23-hardened/node.yaml
```diff
@@ -134,18 +134,18 @@ groups:
         scored: true
 
       - id: 4.1.9
-        text: "Ensure that the kubelet --config configuration file has permissions set to 644 or more restrictive (Automated)"
+        text: "Ensure that the kubelet --config configuration file has permissions set to 600 or more restrictive (Automated)"
```
prachidamle, line 137: Does CIS 1.23 benchmark guide ask for 600 or 644?

galal-hussein: well its actually interesting it check for 644, but it fails because we set it to 600, which is more restrictive it just the comparison is wrong

prachidamle: yeah 600 is stricter anyway, so this is ok

prachidamle (approved):
lgtm

prachidamle:
@galal-hussein  if there is a ticket associated to this change, please add that to the PR so that it is easy to track the change.

mitulshah-suse:
Not sure which ticket this is associated with, but these changes are causing scan failures in rke2 hardened cluster. Please check it. @rishabhmsra please attach failure details... 
@galal-hussein @prachidamle

rishabhmsra:
Rancher version: v2.7-head(35388ea)
Downstream cluster : Hardened RKE2(1-cp, 1-etcd, 1-w) -> `v1.24.4+rke2r1`

**Steps followed:**
- Installed CIS benchmark chart `3.0.0-rc2` with below chart versions:
```
image:
  cisoperator:
    repository: rancher/cis-operator
    tag: v1.0.10-rc2
  securityScan:
    repository: rancher/security-scan
    tag: v0.2.9-rc3
  sonobuoy:
    repository: rancher/mirrored-sonobuoy-sonobuoy
    tag: v0.56.7
```
- Ran  scans using `rke2-cis-1.23-profile-hardened` and `rke2-cis-1.23-profile-permissive` profile

**Result:**
- Getting failures in both permissive and hardened scans:
cis-scans

hardened-failures

permissive-failures

galal-hussein:
@rishabhmsra CIS benchmark 1.23 was decided to run on RKE2 versions >= v1.25.0-rke2r1, since we enable PSA configuration for 1.25 and higher, so for RKE2 versions less than v1.25.0 cis-1.6 should be running

galal-hussein:
@cwayne18 ^^

mitulshah-suse:
@galal-hussein the CIS benchmarks have compatible kubernetes versions associated with them. You can check the following link for the same.
https://github.com/aquasecurity/kube-bench/blob/main/docs/platforms.md#cis-kubernetes-benchmark-support
1.6 scans are not compatible with k8s versions 1.22+ so only the 1.23 scan can be run on these versions.
cc: @prachidamle

galal-hussein:
> @galal-hussein the CIS benchmarks have compatible kubernetes versions associated with them. You can check the following link for the same. https://github.com/aquasecurity/kube-bench/blob/main/docs/platforms.md#cis-kubernetes-benchmark-support 1.6 scans are not compatible with k8s versions 1.22+ so only the 1.23 scan can be run on these versions. cc: @prachidamle

I see, we may need to update our docs then, for the bug I can make the cis-1.23 more generic to work on both PSAs and PSPs so that it can be run on all versions

cwayne18:
cis-1.23 is what we wrote to enable checking for PSA in k8s 1.25, it was never intended to be run against anything < 1.25.

prachidamle:
@cwayne18 @galal-hussein we will have to put up a generic fix to test PSP for k8s versions <1.25 and the PSA fix for k8s versions >= 1.25. 

As per CIS benchmark guide, CIS scan benchmark 1.23 support k8s 1.23 - so the scan need to test for PSP's in this case.

brandond:
Is that page accurate? I don't have a CIS login, but it indicates that the `cis-1.6` profile only supports Kubernetes 1.16-1.18, and the `cis-1.23` profile only supports Kubernetes v1.22-v1.23, which doesn't seem right.

brandond:
If that's the case we should not have been using the `cis-1.6` profile for as long as we were, particularly given that it was apparently only meant for use up to Kubernetes 1.18. What are we supposed to do with `cis-1.23` if it doesn't support 1.24 or 1.25? Why are we just now having issues matching up policy versions to Kubernetes versions when we've apparently been out of sync for basically as long as we supported profiles?

prachidamle:
@brandond yes we have been out of sync w.r.t Kubernetes versions, but it would be good to get this in line with the CIS guides and Kube-bench version support matrix so that we don't have to find fixes for any failures in scans when found on newer kubernetes versions. 

Are there any issues with adding K8s version ranges (min-max) for rke2 scan profiles? 
These changes were recently added to get the version support updated for CIS profiles: https://github.com/rancher/rancher/issues/38439

PR #95, opened 2022-09-21
change pgrep to fix issue for kube-apiserver

mitulshah-suse:
Related to https://github.com/rancher/rancher/issues/32935

### DIFF
--- a/package/run_sonobuoy_plugin.sh
+++ b/package/run_sonobuoy_plugin.sh
@@ -90,14 +90,9 @@ else
   fi
 fi
 
-KUBE_APISERVER_PROC="kube-apiserver"
-if [ ! ${IS_RKE2} ]; then
-  KUBE_APISERVER_PROC="/kube-apiserver"
-fi
-
 # master (no etcd)
 if [[ "${OVERRIDE_BENCHMARK_VERSION}" != "" ]]; then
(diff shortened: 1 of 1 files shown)


prachidamle (commented):
Can we run some tests based on this ticket https://github.com/rancher/rancher/issues/26598 to ensure we dont hit problems?

mitulshah-suse:
> Can we run some tests based on this ticket [rancher/rancher#26598](https://github.com/rancher/rancher/issues/26598) to ensure we dont hit problems?

I created dummy processes on worker node with arguments --kube-apiserver. This was causing failures in existing setup.
The fix ensured that such arguments will not interfere with cis scans. 
The only scenario in which it will fail is if there are binaries named kube-apiserver running on the worker node, which normally would not be the case.

prachidamle (commented):
ok, can we please verify CIS scan tests for all the cluster types rke1/rk2/k3s are passing with the change

mitulshah-suse:
> ok, can we please verify CIS scan tests for all the cluster types rke1/rk2/k3s are passing with the change

yes, i verified for all rke1/rke2/k3s as part of the test..

prachidamle (dismissed):

mitulshah-suse:
For some reason the drone check is not getting triggered and the merging is blocked on that. I tried updating the PR again to trigger the check, but did not help.
@prachidamle Can you please advise? If you have permission, maybe you can merge it as well.

prachidamle (approved):

PR #88, opened 2022-07-05
Fix scan failures after updating kube-bench to v0.6.8 and Go version to 1.17

rayandas:
Related: https://github.com/rancher/cis-operator/issues/156
NOTE: Need to rebase after the following PRs gets merged https://github.com/rancher/security-scan/pull/85, and https://github.com/rancher/security-scan/pull/86
- K3s related fixes for all the profiles (all the fixed are pretty much same):
  - `2.2` and `2.5`: added `bin_op = or`. Some nodes that doesn't have ETCD, it'll thorow errors so it's helpful in cases like this.
  - `1.1.15`, `1.1.16`, : added the schedulerkubeconfig path `/var/lib/rancher/k3s/server/cred/scheduler.kubeconfig`
  - `1.1.17`: added the controllermanagerkubeconfig path `/var/lib/rancher/k3s/server/cred/controller.kubeconfig`
  - `1.1.19`: corrected PKI directory
  - `1.2.2`, `1.2.13`, `1.2.14`, `1.2.15`, `1.2.17`, `1.2.26`: added `-v grep` it's basically grepping the test item using the result of previous grep.
  - (k3s 1.6, 1.20, 1.23 permissive): `1.2.19`, `1.2.20`, `1.2.21`, `1.2.22`, `1.2.23`, `1.2.24`, `1.2.25`: Skipped because these are hardened tests and it'll fail in permissive profiles.
  - `4.2.1`: Modified the script a bit which will hardcode anonymous-auth to false when it doesn't find kube-apiserver on a node (when the node has only etcd role or worker role)
  - `4.2.2`: Modified the script a bit which will hardcode authorization-mode when it doesn't find kube-apiserver on a node (when the node has only etcd role or worker role)
  - `4.2.3`: Modified the script a bit which will hardcode client-ca-file when it doesn't find kube-apiserver on a node (when the node has only etcd role or worker role)
  - `1.1.11`: moved this test inside etcd.yaml for k3s 1.6 permissive profile cz 1.1.11 test is etcd based.

- RKE1 related fixes for all the profiles(all the fixed are pretty much same):
  - `1.1.11`, `1.1.12`: check only the permission of /node/var/lib/etcd
  - `1.1.20`, `1.1.21`: move the .pem path inside single quotes, else the command will throw error
  - `1.1.19`: edit the PKI dir to /node/etc/kubernetes/ssl
  - `4.1.5`, `4.1.6`, : add /node before $kubeletkubeconfig
[...]

PR #96, opened 2022-10-10
Fix for K3s hardened profile test failures.

rayandas:
Related: https://github.com/rancher/cis-operator/issues/169
This PR fixes  `4.2.1`, `4.2.2` and `4.2.3` test failures for `k3s-cis-1.6-hardened`, `k3s-cis-1.20-hardened` and `k3s-cis-1.23-hardened` profiles.

### DIFF
--- a/package/cfg/k3s-cis-1.20-hardened/node.yaml
+++ b/package/cfg/k3s-cis-1.20-hardened/node.yaml
@@ -194,7 +194,7 @@ groups:
     checks:
       - id: 4.2.1
         text: "Ensure that the anonymous-auth argument is set to false (Automated)"
-        audit: "journalctl -D /var/log/journal -u k3s | grep 'Running kube-apiserver' | tail -n1 | grep 'anonymous-auth' | grep -v grep"
+        audit: '/bin/sh -c ''if test $(journalctl -D /var/log/journal -u k3s | grep "Running kube-apiserver" | wc -l) -gt 0; then journalctl -D /var/log/journal -u k3s | grep "Running kube-apiserver" | tail -n1 | grep "anonymous-auth" | grep -v grep; else echo "--anonymous-auth=false"; fi'' '
         tests:
           test_items:
             - flag: "--anonymous-auth"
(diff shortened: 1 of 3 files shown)


package/cfg/k3s-cis-1.20-hardened/node.yaml
```diff
@@ -194,7 +194,7 @@ groups:
     checks:
       - id: 4.2.1
         text: "Ensure that the anonymous-auth argument is set to false (Automated)"
-        audit: "journalctl -D /var/log/journal -u k3s | grep 'Running kube-apiserver' | tail -n1 | grep 'anonymous-auth' | grep -v grep"
+        audit: '/bin/sh -c ''if test $(journalctl -D /var/log/journal -u k3s | grep "Running kube-apiserver" | wc -l) -gt 0; then journalctl -D /var/log/journal -u k3s | grep "Running kube-apiserver" | tail -n1 | grep "anonymous-auth" | grep -v grep; else echo "--anonymous-auth=false"; fi'' '
```
prachidamle, line 197: Could you please explain this fix? we seem to check for the same statement in two ways?

rayandas: So this audit command will hardcode anonymous-auth to false when it doesn't find kube-apiserver on a node after checking  kube-apiserver is running or not.

`if test $(journalctl -D /var/log/journal -u k3s | grep "Running kube-apiserver" | wc -l) -gt 0` this is to check if kube-apiserver is running or not. if if it's running then it'll perform `journalctl -D /var/log/journal -u k3s | grep "Running kube-apiserver" | tail -n1 | grep "anonymous-auth" | grep -v grep;` if not then it will hard-code anonymous-auth to false. 
It was implemented earlier for K3s permissive profile in this PR https://github.com/rancher/security-scan/pull/88

prachidamle (approved):

prachidamle:
Can someone in K3s team also review this @cwayne18

galal-hussein (approved):

rayandas:
Merging the PR

macedogm:
FYI post approving it.

PR #98, opened 2022-10-11
Fix for custom scans readonly dirs

galal-hussein:
- https://github.com/rancher/cis-operator/issues/162


prachidamle (approved):
lgtm. For QA: We should test the custom benchmark scan usecase with this change.

PR #101, opened 2022-11-02
When checking default service accounts avoid an error with clusterrolebindings

manno:
Error was `jq: error (at <stdin>:1715): Cannot iterate over null (null)`

Additionally, clusterrolebindings don't have a namespace, so we check them only once.


andypitcher:
@manno is this PR still relevant ?

PR #103, opened 2022-12-08
Use bitmask in wherever possible instead of multiple comparison with 'or'

rayandas:
Related issue: https://github.com/rancher/cis-operator/issues/120


prachidamle:
hi @rayandas are all the scan profiles changed here passing the scans?

rayandas:
@prachidamle Yeah I have tested all the changes, the scans are passing. 
I've check kube-bench upstream configs as well, they are not using the multiple comparison with 'or' anymore, they're also using bitmask.

prachidamle (approved):
lgtm

PR #109, opened 2023-02-22
Bump dependencies

pjbgf:
The Go version was updated to 1.19, as it was a requirement for some of the dependencies. The base image `golang:<version>-alpine` was replaced with `registry.suse.com/bci/golang:<version>`.

### Go dependencies (direct)
- github.com/aquasecurity/kube-bench v0.6.12
- github.com/sirupsen/logrus v1.9.0
- github.com/spf13/viper v1.15.0
- github.com/urfave/cli v1.22.12


macedogm (commented):
LGTM

Only recommended to use Golang BCI instead of Alpine. Additionally, I would drop the vendoring dependencies.

macedogm:
@rayandas can you also review this PR, please?

CC @prachidamle @MKlimuszka

rayandas:
@pjbgf @macedogm This is a big change as it's bumping `kube-bench` and `sonobuoy`. Last time we had faced some issues after bumping kube-bench and sonobuoy. So these changes might break things, we might need to update `cis-operator` and then `cis-benchmark charts` as well. And we have to add the new sonobuoy image into rancher mirror images as it's being used [here](https://github.com/rancher/charts/blob/dev-v2.7/packages/rancher-cis-benchmark/charts/values.yaml#L13). 

I don't think it can go into v2.7.2 as we are already past feature complete and bug complete. cc: @prachidamle 

After building security-scan image using these changes, I installed CIS-Benchmark on v2.7.2-rc3 using the new image and ran the scans. It's running infinity. Please refer to the snapshot.

Screenshot from 2023-02-22 20-09-34

rayandas:
Just checked there's a PR up for cis-operator as well. Will review that and test it once by running the scans.

rayandas:
@macedogm @prachidamle @MKlimuszka Should we create an issue for these changes and add a milestone?
There’ll be changes in security-scan, cis-operator and then charts. It would be easy to track.

macedogm:
@rayandas your proposal is nice, but I'm afraid that doing so might take some months and then it will not be shipped with our current release (we would have to wait for the next one).

What about if we only merge now the Go dependencies and Go version bump in this PR, then we open another PR with the other dependencies (kube-bench, sonobuoy, kubectl etc.)? At least we can get some basic CVEs fixed right now, while we have more time to test the other updates that can cause breaking changes in the cis-operator and CIS chart?

@pjbgf WDYT?

rayandas:
Yeah that’s a good idea @macedogm 
But still need @prachidamle ’s approval on this.

smangelkramer:
plz check SURE-5916 - current kubectl is more than outdated (its v1.13 !!).
not usable with an CIS  hardened k8s cluster, because this old kubectl does not support current TLS ciphers.

pjbgf:
Updated the PR to include the Go dependencies. Will follow-up with a PR with the external tools bumping.

macedogm (approved):

pjbgf:
Split the tools upgrade into https://github.com/rancher/security-scan/pull/111.

pjbgf:
@rayandas the issue has now been fixed. I tested both changes (#111 and #109) with `quay.io/paulinhu/security-scan:bump1` the scan is now working again.

rayandas:
Built a security scan image using both the changes in PR 109 and 111 (`hackman91/security-scan:v17`)
Built a cis-operator image using the changes in [PR 184](https://github.com/rancher/cis-operator/pull/184) (`hackman91/cis-operator:v5`)
While installing the CIS chart added these images and sonobuoy upstream image (`sonobuoy/sonobuoy:v0.56.15`) and kubectl latest image that rancher's registry have (`rancher/kubectl:tag: v1.25.6`) then installed the chart and ran the scan. 
The scans are passing.

new-scan

The changes looks good to me. We might have to merge security scan PRs 109 and 111 then create a tag and merge cis-operator PR 184 and create a tag then update the charts. 
But need @prachidamle 's approval before doing this.

pjbgf:
Thank you @rayandas. :bow:

rayandas:
@macedogm Is it possible to update this PR as well with latest version of kube-bench? It still uses 0.6.11 in the go.mod
I don't have access to push commits here. Then we're clear with merging #111 and #109 just after a sanity testing and Prachi's approval.

macedogm (approved):

prachidamle (approved):

PR #118, opened 2023-04-11
Add RKE2 CIS 1.24 Configs

rayandas:
Related: https://github.com/rancher/rancher/issues/41886

This PR adds below configs:

- rke2-cis-1.24-permissive
- rke2-cis-1.24-hardened

Depends on Go, kube-bench, sonobuoy, kubectl and other dependency updates. https://github.com/rancher/cis-operator/issues/191

### DIFF
--- a/package/cfg/rke2-cis-1.24-hardened/policies.yaml
+++ b/package/cfg/rke2-cis-1.24-hardened/policies.yaml
@@ -1,6 +1,6 @@
 ---
 controls:
-version: "cis-1.25"
+version: 1.24
 id: 5
 text: "Kubernetes Policies"
 type: "policies"
@@ -42,14 +42,21 @@ groups:
         scored: false
 
       - id: 5.1.5
-        text: "Ensure that default service accounts are not actively used. (Manual)"
-        type: "manual"
+        text: "Ensure that default service accounts are not actively used. (Automated)"
+        audit: check_for_default_sa.sh
+        tests:
+          test_items:
+            - flag: "true"
+              compare:
+                op: eq
+                value: "true"
+              set: true
         remediation: |
           Create explicit service accounts wherever a Kubernetes workload requires specific access
           to the Kubernetes API server.
           Modify the configuration of each default service account to include this value
           automountServiceAccountToken: false
-        scored: false
+        scored: true
 
       - id: 5.1.6
         text: "Ensure that Service Account Tokens are only mounted where necessary (Manual)"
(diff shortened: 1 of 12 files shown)


andypitcher (approved):

package/cfg/rke2-cis-1.24-hardened/policies.yaml
```diff
@@ -0,0 +1,339 @@
...
+                 value: "MustRunAsNonRoot"
+               set: true
+        remediation: |
+          Add policies to each namespace in the cluster which has user workloads to restrict the
+          admission of privileged containers.
+        scored: true
+
+      - id: 5.2.3
+        text: "Minimize the admission of containers wishing to share the host process ID namespace (Automated)"
+        audit: "kubectl get psp -o json | jq .items[] | jq -r 'select((.spec.hostPID == null) or (.spec.hostPID == false))' | jq .metadata.name | wc -l | xargs -I {} echo '--count={}'"
```
dereknola, line 111: These are no longer valid checks for RKE2 v1.25+, which have moved to PSAs. CIS-1.24 was AFAIK supposed to be changed to help validate only v1.25+ versions of both k3s and rke2.

dereknola (changes requested):

package/cfg/rke2-cis-1.24-hardened/policies.yaml
rayandas: @dereknola so should we keep these checks as manual checks([as it is in kube-bench](https://github.com/aquasecurity/kube-bench/blob/main/cfg/cis-1.24/policies.yaml))?
cc: @andypitcher @JonCrowther

rayandas: I think on CIS 1.24, these checks will be present and from CIS 1.25 these will be removed. 
@andypitcher

dereknola: Sure. We can convert them to manual for now.

rayandas: @dereknola No actually we can't convert them to manual, the current behaviour (pass on K8s 1.24 cluster and fail on 1.25 cluster) because CIS 1.24 profiles are supposed to be run on K8s 1.24 clusters and as 1.24 supports PSP these checks will pass. 
On K8s 1.25 cluster these checks won't work as expected because PSP was removed. So when CIS 1.25 Profiles will be released, these checks will be marked as manual because CIS 1.25 profiles can't be run on K8s 1.24 cluster.
cc: @andypitcher

rayandas: FYI @andypitcher is working on adding CIS 1.25 profiles. When that's released, CIS 1.24 profiles will no longer be used to scan K8s 1.25 clusters.

dereknola (approved):

PR #121, opened 2023-05-23
Fix CIS test 1.2.14 for K3s Hardened profile

rayandas:
Related issue: https://github.com/rancher/rancher/issues/41642

Snapshot of the scan on v1.26 K3s hardened cluster:

Screenshot from 2023-05-23 18-06-32

- [ ] TODO: Tag `v0.2.12-rc2` when the PR is merged.

### DIFF
--- a/package/cfg/k3s-cis-1.23-hardened/master.yaml
+++ b/package/cfg/k3s-cis-1.23-hardened/master.yaml
@@ -537,7 +537,7 @@ groups:
 
       - id: 1.2.14
         text: "Ensure that the admission control plugin ServiceAccount is set (Automated)"
-        audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1 | grep 'ServiceAccount'"
+        audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1 | grep -v grep"
         tests:
           bin_op: or
           test_items:


package/cfg/k3s-cis-1.23-hardened/master.yaml
```diff
@@ -537,7 +537,7 @@ groups:
 
       - id: 1.2.14
         text: "Ensure that the admission control plugin ServiceAccount is set (Automated)"
-        audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1 | grep 'ServiceAccount'"
+        audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1 | grep -v grep"
```
prachidamle, line 540: Can you explain what we are doing here using grep -v grep?

brandond: yeah this doesn't make sense to me either. I see this sometimes when grepping through ps output to hide the grep command itself, but that should be necessary yere.

rayandas: @prachidamle @brandond The test was grepping `ServiceAccount` earlier and as it wasn't present in the filtered log entry output hence the output was null or empty. That is why the test was failing.

When using `grep -v grep`  it excludes the grep command itself from the output. and the tests mentioned is to be performed based on the output of the audit command. 

Now as per my understanding, `grep -v grep` works this way: when using the grep command to filter log entries, it searches for lines that match a specific pattern and prints them as output. However, if the pattern being searched (grep) is present in the lines of the log itself, it will also be included in the output. In this case, the grep command used in the audit section is part of the pipeline and will appear in the log entry being retrieved.

To avoid including the grep command line itself in the output, the `grep -v grep` command is used. The -v option in grep -v means "invert match" which excludes lines containing the specified pattern. In this case, it excludes any lines that contain the word "grep" from the output.

By using grep -v grep, only the relevant log entry that contains the phrase "Running kube-apiserver" will be selected, and the grep command used in the pipeline will be filtered out from the output.

So, the purpose of grep -v grep is to ensure that the output includes only the relevant log entry while excluding the grep command itself.

brandond: > When using `grep -v grep` it excludes the grep command itself from the output. and the tests mentioned is to be performed based on the output of the audit command.

There is no reason for the word `grep` to appear in the output in the first place. This command is grepping through logs, not process names, grep will never occur.

What is the intended behavior here? Are we searching for a specific string in the logs? If so, you've just broken that check by matching anything that does NOT contain the word grep, which is not a proper validation of the control.

rayandas: My bad, in this case we can remove the grep part. This will be the audit command. I just tested it and it's working.
```
audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1"

Comment (brandond) [reply package/cfg/k3s-cis-1.23-hardened/master.yaml:540]: If you do that you're just checking to see if kube-apiserver was run, which doesn't seem correct either?

Comment (rayandas) [reply package/cfg/k3s-cis-1.23-hardened/master.yaml:540]: I don't think that it's incorrect. 
```
audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1"
```
It will retrieve the most recent log entry related to the `kube-apiserver` process from the k3s service logs. 
```
 tests:
      bin_op: or
      test_items:
        - flag: "--disable-admission-plugins"
          compare:
            op: nothave
            value: "ServiceAccount"
        - flag: "--disable-admission-plugins"
          set: false
```
And then the above test items performs the checks on the output that it got after running the audit command.

Comment (prachidamle) [reply package/cfg/k3s-cis-1.23-hardened/master.yaml:540]: @rayandas Maybe can you paste the the output that it got after running the audit command to see what the log looks like?

Comment (prachidamle) [reply package/cfg/k3s-cis-1.23-hardened/master.yaml:540]: Also, with this change does the CIS scan work for previous K8s versions since now we are not grepping anymore for "Serviceaccount"?

Comment (brandond) [reply package/cfg/k3s-cis-1.23-hardened/master.yaml:540]: If there is additional filtering of the output, then that looks good to me.

Comment (rayandas) [reply package/cfg/k3s-cis-1.23-hardened/master.yaml:540]: @prachidamle @brandond I have pasted the outputs of three different audit commands (below) here on the therpad: https://pad.riseup.net/p/scan-outputs
```
1. audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1 | grep -v grep"
2. audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1"
3. audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1 | grep 'ServiceAccount'"
```
Note: Output of the command 1 and 2 are identical.

I have tested running scans on 1.24 and 1.25 K3s clusters as well when the audit command is ``audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1 | grep -v grep"`` and the scan 1.2.14 passed. 

Though I haven't got a chance to test with audit command ``audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1"`` on 1.24 and 1.25 cluster but tested it on 1.26 cluster and it passed. Also I am certain that it will pass on 1.24 and 1.25 clusters as well. I can confirm it on Monday after testing on those clusters.

brandond: `audit: "journalctl -D /var/log/journal  -u k3s | grep 'Running kube-apiserver' | tail -n1"` sounds good to me.

prachidamle (approved):
Filtering of the output lgtm

PR #131, opened 2023-06-20
Set `SKIP` to `State` instead of `NA` when `Type` is `SKIP`

masap:


masap:
Close for now. After this PR was merged to leodotcloud branch, I will resend this.

PR #141, opened 2023-07-10
Use Drone's plugin for GitHub releases

macedogm:
Use Drone's plugin for GitHub releases instead of a non-official image, which is tied to a user's DockerHub repo.

Note: I can't test if the entire config is right. This is only possible when we cut the next -rc.


macedogm:
Closed by mistake and can't reopen it.

PR #146, opened 2023-07-24
Fix k3s-cis-1.7 permissive node.yaml 4.2.6 --make-iptables-util-chains

andypitcher:
Related to this BUG: [[BUG] Test 4.2.6 in CIS scan fails on K3S cluster](https://github.com/rancher/rancher/issues/42138).
The `type: skip` is now implemented in _k3s-cis-1.7-permissive_ for `4.2.6 --make-iptables-util-chains` to match with the previous k3s-cis-permissive profiles (1.5-1.24).


rayandas (approved):
LGTM

PR #148, opened 2023-07-24
Fix regression: missing upstream tests

pjbgf:
- Fixes regression introduced by https://github.com/rancher/security-scan/commit/1836d3c10467dccffe9cd6f266d9722c5ce2cbff whereby the kube-bench cfg files were misplaced.
- Adds tests to avoid future regressions.


andypitcher (approved):
LGTM, approving.

PR #152, opened 2023-08-11
Fix k3s cis 1.7 check errors

JonCrowther:
Addresses https://github.com/rancher/rancher/issues/42384

Fixes 3 errors in both k3s-cis-1.7-hardened and permissive

### DIFF
--- a/package/cfg/k3s-cis-1.7-hardened/master.yaml
+++ b/package/cfg/k3s-cis-1.7-hardened/master.yaml
@@ -151,7 +151,7 @@ groups:
       - id: 1.1.10
         text: "Ensure that the Container Network Interface file ownership is set to root:root (Manual)"
         audit: |
-          ps -ef | grep $kubeletbin | grep -- --cni-conf-dir | sed 's%.*cni-conf-dir[= ]\([^ ]*\).*%\1%' | xargs -I{} find {} -mindepth 1 | xargs --no-run-if-empty stat -c %U:%G
+          ps -ef | grep $kubeletbin | grep -- --cni-conf-dir | echo "/etc/cni/net.d" | sed 's%.*cni-conf-dir[= ]\([^ ]*\).*%\1%' | xargs --no-run-if-empty stat -c %U:%G
           find /var/lib/cni/networks -type f 2> /dev/null | xargs --no-run-if-empty stat -c %U:%G
         use_multiple_values: true
         tests:
@@ -285,7 +285,7 @@ groups:
 
       - id: 1.1.19
         text: "Ensure that the Kubernetes PKI directory and file ownership is set to root:root (Automated)"
-        audit: "find /var/lib/rancher/k3s/server/tls | xargs stat -c %U:%G"
+        audit: "stat -c %U:%G /var/lib/rancher/k3s/server/tls"
         use_multiple_values: true
         tests:
           test_items:
--- a/package/cfg/k3s-cis-1.7-permissive/master.yaml
+++ b/package/cfg/k3s-cis-1.7-permissive/master.yaml
@@ -151,7 +151,7 @@ groups:
       - id: 1.1.10
         text: "Ensure that the Container Network Interface file ownership is set to root:root (Manual)"
         audit: |
-          ps -ef | grep $kubeletbin | grep -- --cni-conf-dir | sed 's%.*cni-conf-dir[= ]\([^ ]*\).*%\1%' | xargs -I{} find {} -mindepth 1 | xargs --no-run-if-empty stat -c %U:%G
+          ps -ef | grep $kubeletbin | grep -- --cni-conf-dir | echo "/etc/cni/net.d" | sed 's%.*cni-conf-dir[= ]\([^ ]*\).*%\1%' | xargs --no-run-if-empty stat -c %U:%G
           find /var/lib/cni/networks -type f 2> /dev/null | xargs --no-run-if-empty stat -c %U:%G
         use_multiple_values: true
         tests:
@@ -285,7 +285,7 @@ groups:
 
       - id: 1.1.19
         text: "Ensure that the Kubernetes PKI directory and file ownership is set to root:root (Automated)"
-        audit: "find /var/lib/rancher/k3s/server/tls | xargs stat -c %U:%G"
+        audit: "stat -c %U:%G /var/lib/rancher/k3s/server/tls"
         use_multiple_values: true
         tests:
           test_items:
--- a/package/cfg/k3s-cis-1.7-hardened/node.yaml
+++ b/package/cfg/k3s-cis-1.7-hardened/node.yaml
@@ -46,7 +46,6 @@ groups:
           bin_op: or
           test_items:
             - flag: "permissions"
-              set: true
               compare:
                 op: bitmask
                 value: "600"
--- a/package/cfg/k3s-cis-1.7-permissive/node.yaml
+++ b/package/cfg/k3s-cis-1.7-permissive/node.yaml
@@ -46,7 +46,6 @@ groups:
           bin_op: or
           test_items:
             - flag: "permissions"
-              set: true
               compare:
                 op: bitmask
                 value: "600"


Priyashetty17 (approved):

package/cfg/k3s-cis-1.7-hardened/master.yaml
```diff
@@ -151,7 +151,7 @@ groups:
       - id: 1.1.10
         text: "Ensure that the Container Network Interface file ownership is set to root:root (Manual)"
         audit: |
-          ps -ef | grep $kubeletbin | grep -- --cni-conf-dir | sed 's%.*cni-conf-dir[= ]\([^ ]*\).*%\1%' | xargs -I{} find {} -mindepth 1 | xargs --no-run-if-empty stat -c %U:%G
+          ps -ef | grep $kubeletbin | grep -- --cni-conf-dir | echo "/etc/cni/net.d" | sed 's%.*cni-conf-dir[= ]\([^ ]*\).*%\1%' | xargs --no-run-if-empty stat -c %U:%G
```
KevinJoiner, line 154: What is the point of
 ``` bash
ps -ef | grep $kubeletbin | grep -- --cni-conf-dir | echo "/etc/cni/net.d"` 
```
`echo` does not use stdin so this statement is equivalent to 
``` bash
echo "/etc/cni/net.d"`
```

The problem here is that $kubectlbin is unset. I am not sure what version of bash is being used but something like this may be what we want
``` bash
ps -ef | kubeletbin=${$kubeletbin:="<DEFAULT_KUBECTL_PATH>"} grep $kubeletbin | grep -- --cni-conf-dir | sed 's%.*cni-conf-dir[= ]\([^ ]*\).*%\1%' | xargs -I{} find {} -mindepth 1 | xargs --no-run-if-empty stat -c %U:%G
````

File: package/cfg/k3s-cis-1.7-hardened/master.yaml [hunk 2]
```diff
@@ -285,7 +285,7 @@ groups:
 
       - id: 1.1.19
         text: "Ensure that the Kubernetes PKI directory and file ownership is set to root:root (Automated)"
-        audit: "find /var/lib/rancher/k3s/server/tls | xargs stat -c %U:%G"
+        audit: "stat -c %U:%G /var/lib/rancher/k3s/server/tls"
```
KevinJoiner, line 288: This changes the desired check only to check the directory and not all recursive files. The original command is correct. We just need to update the test_items configuration to compare the output and validate that the only strings returned are `root:root` 
So this should pass
```
root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:roo
```

This should not pass.
```
root:root root:root NOT_THE_ROOT_USER:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:root root:roo
```
This can be done with regex

package/cfg/k3s-cis-1.7-permissive/master.yaml
```diff
@@ -151,7 +151,7 @@ groups:
       - id: 1.1.10
         text: "Ensure that the Container Network Interface file ownership is set to root:root (Manual)"
         audit: |
-          ps -ef | grep $kubeletbin | grep -- --cni-conf-dir | sed 's%.*cni-conf-dir[= ]\([^ ]*\).*%\1%' | xargs -I{} find {} -mindepth 1 | xargs --no-run-if-empty stat -c %U:%G
+          ps -ef | grep $kubeletbin | grep -- --cni-conf-dir | echo "/etc/cni/net.d" | sed 's%.*cni-conf-dir[= ]\([^ ]*\).*%\1%' | xargs --no-run-if-empty stat -c %U:%G
```
KevinJoiner, line 154: Same issue as above

package/cfg/k3s-cis-1.7-permissive/master.yaml
```diff
@@ -285,7 +285,7 @@ groups:
 
       - id: 1.1.19
         text: "Ensure that the Kubernetes PKI directory and file ownership is set to root:root (Automated)"
-        audit: "find /var/lib/rancher/k3s/server/tls | xargs stat -c %U:%G"
+        audit: "stat -c %U:%G /var/lib/rancher/k3s/server/tls"
```
KevinJoiner, line 288: same issue as above

KevinJoiner (changes requested):

package/cfg/k3s-cis-1.7-hardened/master.yaml
andypitcher: @KevinJoiner the kube-bench variables such as `$kubeletbin` are defined in config.yaml which can be set globally or locally per profile.

Path for "global"/ default : `security-scan/package/cfg/config.yaml`
Path for local: `security-scan/package/cfg/k3s-cis-1.7-permissive/config.yaml`

It would be better to set the variables for each profile's `config.yaml` (locally), instead of hard coding the values in the `audit` comands, to stay align with kube-bench's standards.

For instance in `k3s-cis-1.7-permissive`, the current local config map sets` $kubeletbin` to `containerd`, as shown [here](https://github.com/rancher/security-scan/blob/bb9f870e95d8d709ad090bc2832ba3bc6ee73c8a/package/cfg/k3s-cis-1.7-permissive/config.yaml#L35).
If `containerd` is not found, the other bins might be used by kube-bench as show [here](https://github.com/rancher/security-scan/blob/bb9f870e95d8d709ad090bc2832ba3bc6ee73c8a/package/cfg/config.yaml#L109).

There is an open issue, to go over all possible variables for each distro and assign their values, it would be great if we could fix all these values here.

- https://github.com/rancher/cis-operator/issues/211

As for this specific `audit`, we should test the [original](https://github.com/aquasecurity/kube-bench/blob/d70459b77c28adf7663abbe87458ab109665cb70/cfg/cis-1.7/master.yaml#L133C1-L133C1) one from kube-bench with the right `$kubeletbin` value (set in config.yaml).

package/cfg/k3s-cis-1.7-hardened/master.yaml
KevinJoiner: After taking another look at this you should be able to leverage this script to accomplish our goal using [check_files_owner_dir.sh](https://github.com/rancher/security-scan/blob/master/package/helper_scripts/check_files_owner_in_dir.sh) similar to what is done in this [test](https://github.com/rancher/security-scan/blob/bb9f870e95d8d709ad090bc2832ba3bc6ee73c8a/package/cfg/rke-cis-1.7-permissive/etcd.yaml#L45-L59)

package/cfg/k3s-cis-1.7-hardened/master.yaml
andypitcher: Sounds good, not 100% sure if we could transfer all the parameters from the global `config.yaml` to the local ones. It should work for all the components variables, though for `target_mapping` and the default variables I would keep the global one intact.

JonCrowther: Just to clarify, because there's an open issue, I should revert this to the original command for now and the bug will be fixed when https://github.com/rancher/cis-operator/issues/211 gets resolved? Or is there a correct value I should change it to now?

package/cfg/k3s-cis-1.7-hardened/config.yaml
```diff
@@ -32,7 +32,9 @@ master:
 
     kubelet:
       bins:
-        - containerd
+        - "hyperkube kubelet"
+        - "kubelet"
+        - "containerd"
```
KevinJoiner, line 37: I would move containerd to the top to help preserve any previous behavior.

KevinJoiner (changes requested):

KevinJoiner (approved):

PR #166, opened 2023-09-29
Dockerfile multistage with bci-micro

andypitcher:
Parent issue: https://github.com/rancher/security-team/issues/387

This PR covers the following:

- Have a multi-stage approach
- Use [bci-micro](https://registry.suse.com/bci/bci-micro-15sp5/index.html) as final image and remove unused packages.

Tested with RKE2, the scan is working and delivers the same results:

image

Note: The final image size can't be smaller than ~280MB even if we're using bci-micro here. Indeed, the number of binaries and their dependencies increases it. This said here are some advantages using micro here:

1. It reduces the attack surface, for example in bci-base `zypper` is here by default. This might help an attacker to install other artifacts onto the container.
2. It reduces the number of potential vulnerabilities. After the build, the number of binaries in /usr/bin is equal to `403` when using `bci-base` and `354` when using `bci-micro`.

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -31,7 +31,7 @@ RUN mkdir -p /chroot/etc/kube-bench/ && \
     tar xvz -C /chroot/etc/kube-bench/ --strip-components=1 "kube-bench-${kube_bench_version}/cfg"
 
 ## OS binaries to run kube-bench audit commands
-RUN zypper --installroot /chroot -n --gpg-auto-import-keys in --no-recommends findutils tar jq gawk diffutils procps systemd gzip  && \
+RUN zypper --installroot /chroot -n --gpg-auto-import-keys in --no-recommends findutils tar jq gawk diffutils procps systemd gzip curl  && \
     zypper --installroot /chroot clean -a && \
     rm -rf /chroot/var/cache/zypp/* /chroot/var/log/zypp/*
 
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,29 +1,57 @@
-FROM registry.suse.com/bci/bci-base:15.5
+# Temporary build stage
+FROM registry.suse.com/bci/bci-base:15.5 as builder
 
-ARG kube_bench_version=0.6.12
+# Define build arguments
+ARG kube_bench_version=0.6.17
 ARG sonobuoy_version=0.56.16
 ARG kubectl_version=1.28.0
 ARG ARCH
 
-RUN zypper --non-interactive update \
-    && zypper --non-interactive install \
-    systemd \
-    curl \
-    jq \
-    tar \
-    awk \
-    gzip
+# Install build dependencies
+RUN zypper --non-interactive update && zypper --non-interactive install jq awk tar systemd
+
+# Install kubectl
 RUN curl -Lo ./kubectl "https://storage.googleapis.com/kubernetes-release/release/v${kubectl_version}/bin/linux/${ARCH}/kubectl" && \
     chmod +x ./kubectl && \
     mv ./kubectl /usr/local/bin/
+
+# Install Sonobuoy
 RUN curl -sLf "https://github.com/vmware-tanzu/sonobuoy/releases/download/v${sonobuoy_version}/sonobuoy_${sonobuoy_version}_linux_${ARCH}.tar.gz" | tar -xvzf - -C /usr/bin sonobuoy
+
+# Install kube-bench
 RUN curl -sLf "https://github.com/aquasecurity/kube-bench/releases/download/v${kube_bench_version}/kube-bench_${kube_bench_version}_linux_${ARCH}.tar.gz" | tar -xvzf - -C /usr/bin
 
-# Copy the files within /cfg straight from the immutable GitHub source to /etc/kube-bench/cfg/.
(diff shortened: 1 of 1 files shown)


pjbgf (dismissed):

pjbgf:
[Diving](https://github.com/wagoodman/dive) in `rancher/security-scan:v0.2.13`, the binaries we download are around 150mb:
image

We could potentially go to `registry.suse.com/bci/bci-busybox:15.5` which is ~12mb. So in total, we should be able to get to sub 200MB.

package/Dockerfile
```diff
@@ -1,29 +1,47 @@
...
+RUN curl -sLf "https://github.com/aquasecurity/kube-bench/releases/download/v${kube_bench_version}/kube-bench_${kube_bench_version}_linux_${ARCH}.tar.gz" | tar -xvzf - -C /chroot/usr/bin
+
+## Copy the files within /cfg straight from the immutable GitHub source to /etc/kube-bench/cfg/ into micro
+RUN mkdir -p /chroot/etc/kube-bench/ && \
     curl -sLf "https://github.com/aquasecurity/kube-bench/archive/refs/tags/v${kube_bench_version}.tar.gz" | \
-    tar xvz -C /etc/kube-bench/ --strip-components=1 "kube-bench-${kube_bench_version}/cfg"
+    tar xvz -C /chroot/etc/kube-bench/ --strip-components=1 "kube-bench-${kube_bench_version}/cfg"
+
+## OS binaries to run kube-bench audit commands
+RUN zypper --installroot /chroot -n --gpg-auto-import-keys in --no-recommends findutils tar jq gawk diffutils procps systemd gzip curl  && \
```
macedogm, line 34:
```suggestion
RUN zypper --installroot /chroot -n in --no-recommends findutils tar jq gawk diffutils procps systemd gzip curl  && \
```

macedogm: `--gpg-auto-import-keys` isn't needed, because we aren't adding new repos in Zypper.

macedogm (changes requested):

package/Dockerfile
andypitcher: @macedogm removing the flag `--gpg-auto-import-keys` makes the zypper command to fail. It asks for a manual validation:

```
 > [builder 7/7] RUN zypper --installroot /chroot -n in --no-recommends findutils tar jq gawk diffutils procps systemd gzip curl  &&     zypper --installroot /chroot clean -a &&     rm -rf /chroot/var/cache/zypp/* /chroot/var/log/zypp/*:
#0 0.306 Refreshing service 'container-suseconnect-zypp'.
#0 0.317 Retrieving repository 'SLE_BCI' metadata [......
#0 3.239
#0 3.239 New repository or package signing key received:
#0 3.239
#0 3.239   Repository:       SLE_BCI
#0 3.239   Key Fingerprint:  FEAB 5025 39D8 46DB 2C09 61CA 70AF 9E81 39DB 7C82
#0 3.239   Key Name:         SuSE Package Signing Key <build@suse.de>
#0 3.239   Key Algorithm:    RSA 2048
#0 3.239   Key Created:      Mon Sep 21 08:21:47 2020
#0 3.239   Key Expires:      Fri Sep 20 08:21:47 2024
#0 3.239   Rpm Name:         gpg-pubkey-39db7c82-5f68629b
#0 3.239
#0 3.239
#0 3.239
#0 3.239     Note: Signing data enables the recipient to verify that no modifications occurred after the data
#0 3.239     were signed. Accepting data with no, wrong or unknown signature can lead to a corrupted system
#0 3.239     and in extreme cases even to a system compromise.
#0 3.239
#0 3.239     Note: A GPG pubkey is clearly identified by its fingerprint. Do not rely on the key's name. If
#0 3.239     you are not sure whether the presented key is authentic, ask the repository provider or check
#0 3.239     their web site. Many providers maintain a web page showing the fingerprints of the GPG keys they
#0 3.239     are using.
#0 3.239
#0 3.239 Do you want to reject the key, trust temporarily, or trust always? [r/t/a/?] (r): r
#0 3.245 error]
```

pjbgf (dismissed):

macedogm (dismissed):

rayandas:
@andypitcher Could you please resolve the conflicts?

andypitcher:
> @andypitcher Could you please resolve the conflicts?

Done. also replaced kubectl download URL from storage.googleapis.com to dl.k8s.io as per https://github.com/kubernetes/k8s.io/issues/2396

macedogm (dismissed):

package/Dockerfile
```diff
@@ -1,29 +1,44 @@
...
+    zypper --installroot /chroot clean -a && \
+    rm -rf /chroot/var/cache/zypp/* /chroot/var/log/zypp/*
+
+# Main stage using bco-mirco as the base image
```
pjbgf, line 35: nit:
```suggestion
# Main stage using bco-micro as the base image
```

pjbgf (dismissed):

pjbgf (approved):

MKlimuszka:
Infracloud is testing this.

pjbgf:
@MKlimuszka it may be worth testing this together with https://github.com/rancher/cis-operator/pull/232.

PR #179, opened 2023-12-13
Add CIS-1.8 for RKE, RKE2 and K3s

rayandas:
Related: https://github.com/rancher/rancher/issues/43159 and https://github.com/rancher/rancher/issues/43710

This PR adds CIS-1.8 configs for RKE, RKE2 and K3s. 

> NOTE: There are not much changes since CIS-1.7 except:
> - Test 1.2.16 (master.yaml) -  Ensure that the --secure-port argument is not set to 0 - Note This recommendation is obsolete and will be deleted per the consensus process (Manual)" REMOVED in CIS-1.8, numbering changed accordingly (1.2.16 becomes Ensure that the --profiling argument is set to false (Automated))
> - Test 4.1.9 (node.yaml) - "If the kubelet config.yaml configuration file is being used validate permissions set to 600 or more restrictive (Manual)" BECOMES automated in CIS-1.8.
> - Test 4.1.10 (node.yaml) - "If the kubelet config.yaml configuration file is being used validate file ownership is set to root:root (Manual)" BECOMES automated in CIS-1.8.

It can be checked using below command
```
➜  cfg git:(main) ✗ for component in {config.yaml,controlplane.yaml,etcd.yaml,master.yaml,node.yaml,policies.yaml}; do echo "[$component]" && diff cis-1.7/$component cis-1.8/$component | tee -a changelog_$component; done
[config.yaml]
[controlplane.yaml]
3c3
< version: "cis-1.7"
---
> version: "cis-1.8"
[etcd.yaml]
3c3
< version: "cis-1.7"
---
> version: "cis-1.8"
[master.yaml]
3c3
< version: "cis-1.7"
---
> version: "cis-1.8"
564,581d563
<         text: "Ensure that the --secure-port argument is not set to 0 - NoteThis recommendation is obsolete and will be deleted per the consensus process (Manual)"
<         audit: "/bin/ps -ef | grep $apiserverbin | grep -v grep"
<         tests:
<           bin_op: or
<           test_items:
<             - flag: "--secure-port"
<               compare:
<                 op: gt
<                 value: 0
<             - flag: "--secure-port"
<               set: false
<         remediation: |
<           Edit the API server pod specification file $apiserverconf
<           on the control plane node and either remove the --secure-port parameter or
<           set it to a different (non-zero) desired port.
<         scored: false
<
<       - id: 1.2.17
596c578
<       - id: 1.2.18
---
>       - id: 1.2.17
609c591
<       - id: 1.2.19
---
[...]

PR #184, opened 2024-01-12
fix k3s scan profiles

vardhaman22:
related issue: https://github.com/rancher/cis-operator/issues/254, https://github.com/rancher/cis-operator/issues/255, https://github.com/rancher/cis-operator/issues/258, https://github.com/rancher/cis-operator/issues/259, https://github.com/rancher/cis-operator/issues/260 and https://github.com/rancher/cis-operator/issues/261
Note: fixed k3s profiles for k8s v1.24 and above onl

1. k3s scans were not running on worker nodes because in the run_sonobuoy_plugin script, to check the worker node k3s-agent service logs were not checked.
to fix this, i added the k3s-agent service in that check.
```
  if [[ "$(pgrep kubelet | wc -l)" -gt 0 ]] || [[ "$(journalctl -D $JOURNAL_LOG -u k3s -u k3s-agent | grep -m1 'Running kubelet' | wc -l)" -gt 0 ]]; then
```

2. due to the above fix some tests were broken so fixed those test cases 4.1.8, 4.2.4, 4.2.5 and 4.2.6.

2. cis 1.8 hardened config.yaml was fixed in commit https://github.com/rancher/security-scan/pull/179/commits/614996f078444ba7ec1a5b34bb16cec18bb14618 but this exist in the same way in all other profile configuration for k3s so fixed that.

3.  current sed command was only removing one occurrence in a line so added `|g` to it to correctly replace all occurrences of /var/log/journal in the script.

### DIFF
--- a/package/cfg/k3s-cis-1.24-hardened/config.yaml
+++ b/package/cfg/k3s-cis-1.24-hardened/config.yaml
@@ -25,22 +25,22 @@ master:
     bins:
       - containerd
 
-  node:
-    components:
-      - kubelet
-      - proxy
+node:
+  components:
+    - kubelet
+    - proxy
 
-    kubelet:
-      bins:
-        - containerd
-      defaultkubeconfig: /var/lib/rancher/k3s/agent/kubelet.kubeconfig
-      defaultcafile: /var/lib/rancher/k3s/agent/client-ca.crt
+  kubelet:
+    bins:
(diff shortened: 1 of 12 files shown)


pdellamore (approved):

pjbgf (approved):
LGTM

PR #191, opened 2024-02-14
fix 4.1.7 test case for k3s-cis-1.23-permissive profile

vardhaman22:
issue: https://github.com/rancher/cis-operator/issues/262
1. changed `scored: true` to ` scored: false ` for 4.1.7 test case for  k3s-cis-1.23-permissive based on upstream.
2. fix test items for  k3s-cis-1.23-permissive and  k3s-cis-1.23-hardened in test 4.1.7
3. fix file path for certificate file in 4.1.7 test audit in  k3s-cis-1.23-permissive and  k3s-cis-1.23-hardened

### DIFF
--- a/package/cfg/k3s-cis-1.23-permissive/node.yaml
+++ b/package/cfg/k3s-cis-1.23-permissive/node.yaml
@@ -103,49 +103,17 @@ groups:
 
       - id: 4.1.7
         text: "Ensure that the certificate authorities file permissions are set to 644 or more restrictive (Manual)"
-        audit: "stat -c %a /var/lib/rancher/k3s/server/tls/server-ca.crt"
+        audit: "stat -c %a $kubeletcafile"
         tests:
           test_items:
-            - flag: "644"
+            - flag: "permissions"
               compare:
-                op: eq
+                op: bitmask
                 value: "644"
-              set: true
-            - flag: "640"
-              compare:
-                op: eq
-                value: "640"
-              set: true
-            - flag: "600"
-              compare:
-                op: eq
-                value: "600"
-              set: true
-            - flag: "444"
-              compare:
-                op: eq
-                value: "444"
-              set: true
-            - flag: "440"
-              compare:
-                op: eq
-                value: "440"
-              set: true
-            - flag: "400"
-              compare:
-                op: eq
-                value: "400"
-              set: true
-            - flag: "000"
-              compare:
-                op: eq
-                value: "000"
-              set: true
-          bin_op: or
         remediation: |
           Run the following command to modify the file permissions of the
           --client-ca-file chmod 644 <filename>
-        scored: true
+        scored: false
 
       - id: 4.1.8
         text: "Ensure that the client certificate authorities file ownership is set to root:root (Manual)"
--- a/package/cfg/k3s-cis-1.23-hardened/node.yaml
+++ b/package/cfg/k3s-cis-1.23-hardened/node.yaml
@@ -103,45 +103,13 @@ groups:
 
       - id: 4.1.7
         text: "Ensure that the certificate authorities file permissions are set to 644 or more restrictive (Manual)"
-        audit: "stat -c %a /var/lib/rancher/k3s/server/tls/server-ca.crt"
+        audit: "stat -c %a $kubeletcafile"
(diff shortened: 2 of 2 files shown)


andypitcher (changes requested):
The correction LGTM, this check should be manual (scored: False).

However, my recommendation would be to adjust the check based on kube-bench's [one](https://github.com/aquasecurity/kube-bench/blob/3db3f736f8b5423882f22747c51978849d56fded/cfg/cis-1.23/node.yaml#L95). (4.1.8 could be adjusted as well).

```
      - id: 4.1.7
        text: "Ensure that the certificate authorities file permissions are set to 644 or more restrictive (Manual)"
        audit: |
          CAFILE=$(ps -ef | grep kubelet | grep -v apiserver | grep -- --client-ca-file= | awk -F '--client-ca-file=' '{print $2}' | awk '{print $1}' | uniq)
          if test -z $CAFILE; then CAFILE=$kubeletcafile; fi
          if test -e $CAFILE; then stat -c permissions=%a $CAFILE; fi
        tests:
          test_items:
            - flag: "permissions"
              compare:
                op: bitmask
                value: "644"
        remediation: |
          Run the following command to modify the file permissions of the
          --client-ca-file chmod 644 <filename>
        scored: false
```

package/cfg/k3s-cis-1.23-permissive/node.yaml
```diff
@@ -103,53 +103,27 @@ groups:
 
       - id: 4.1.7
         text: "Ensure that the certificate authorities file permissions are set to 644 or more restrictive (Manual)"
-        audit: "stat -c %a /var/lib/rancher/k3s/server/tls/server-ca.crt"
+        audit: |
+          CAFILE=$(ps -ef | grep kubelet | grep -v apiserver | grep -- --client-ca-file= | awk -F '--client-ca-file=' '{print $2}' | awk '{print $1}' | uniq)
+          if test -z $CAFILE; then CAFILE=$kubeletcafile; fi
+          if test -e $CAFILE; then stat -c permissions=%a $CAFILE; fi
```
andypitcher, line 109: The change I proposed might fail for K3s, I recommend you keep the previous `audit` (based on [k3s-cis-1.7-permissive](https://github.com/rancher/security-scan/blob/e8dcabab764ea4915e34af1bdc206965ef25591a/package/cfg/k3s-cis-1.7-permissive/node.yaml#L99C9-L99C74)) but change ONLY the items (as you did here).

```suggestion
        audit: "stat -c %a /var/lib/rancher/k3s/server/tls/server-ca.crt"
```

package/cfg/k3s-cis-1.23-permissive/node.yaml
```diff
@@ -103,53 +103,27 @@ groups:
...
-        scored: true
+        scored: false
 
       - id: 4.1.8
         text: "Ensure that the client certificate authorities file ownership is set to root:root (Manual)"
-        audit: "stat -c %U:%G $kubeletcafile"
+        audit: |
+          CAFILE=$(ps -ef | grep kubelet | grep -v apiserver | grep -- --client-ca-file= | awk -F '--client-ca-file=' '{print $2}' | awk '{print $1}' | uniq)
+          if test -z $CAFILE; then CAFILE=$kubeletcafile; fi
+          if test -e $CAFILE; then stat -c %U:%G $CAFILE; fi
```
andypitcher, line 126: Same as mentioned above, we should actually keep the same way of auditing in 4.1.8 (ps | grep kubelet should be empty in K3s).

```suggestion
        audit: stat -c %U:%G $kubeletcafile
```

andypitcher (changes requested):

package/cfg/k3s-cis-1.23-permissive/node.yaml
vardhaman22: done

package/cfg/k3s-cis-1.23-permissive/node.yaml
vardhaman22: done

andypitcher (approved):
lgtm

pjbgf (approved):

PR #207, opened 2024-04-16
k3s profiles: fix remediation for 1.1.20 and 1.1.21 checks

vardhaman22:
https://github.com/rancher/cis-operator/issues/290, https://github.com/rancher/cis-operator/issues/291 and https://github.com/rancher/cis-operator/issues/292

fix command in remediation text for 1.1.20 and 1.1.21 checks.
existing command when used in shell was giving an error:
```find: missing argument to `-exec'```


andypitcher (approved):
Good catch, approved.

PR #213, opened 2024-06-27
Update bci-micro version

krunalhinguu:
### Updates

- Update `bci-micro` version to `15.6`

### Related

- https://github.com/rancher/rancher/issues/45588

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -39,7 +39,7 @@ COPY --from=golang /etc/zypp/ /chroot/etc/zypp/
 # OS binaries to run kube-bench audit commands.
 # Before removing dependencies here, ensure they are not needed by cis-operator,
 # security-scan or any of the existing kube-bench tests.
-RUN zypper --non-interactive refresh && \
+RUN zypper --non-interactive --releasever=15.6 refresh && \
     zypper --installroot /chroot -n in --no-recommends findutils tar jq gawk diffutils procps systemd gzip curl && \
     zypper --installroot /chroot clean -a && \
     rm -rf /chroot/var/cache/zypp/* /chroot/var/log/zypp/* /chroot/etc/zypp/
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,5 +1,5 @@
 # Platform specific images which must support all target architectures.
-FROM registry.suse.com/bci/bci-micro:15.5 AS final
+FROM registry.suse.com/bci/bci-micro:15.6 AS final
 FROM registry.suse.com/bci/golang:1.22 AS golang
 
 # Builder and xx only need to support the host architecture.
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -3,7 +3,7 @@ FROM registry.suse.com/bci/bci-micro:15.6 AS final
 FROM registry.suse.com/bci/golang:1.22 AS golang
 
 # Builder and xx only need to support the host architecture.
-FROM --platform=$BUILDPLATFORM rancher/mirrored-tonistiigi-xx:1.3.0 as xx
+FROM --platform=$BUILDPLATFORM rancher/mirrored-tonistiigi-xx:1.3.0 AS xx
 FROM --platform=$BUILDPLATFORM registry.suse.com/bci/golang:1.22 AS builder
 
 # Bring xx supporting tools to the builder layer, but as it is at / it


package/Dockerfile
```diff
@@ -39,8 +39,8 @@ COPY --from=golang /etc/zypp/ /chroot/etc/zypp/
 # OS binaries to run kube-bench audit commands.
 # Before removing dependencies here, ensure they are not needed by cis-operator,
 # security-scan or any of the existing kube-bench tests.
-RUN zypper --non-interactive refresh && \
-    zypper --installroot /chroot -n in --no-recommends findutils tar jq gawk diffutils procps systemd gzip curl && \
+RUN zypper --non-interactive --gpg-auto-import-keys refresh && \
```
pjbgf, line 42: Do we need to auto import keys? Ideally we would import the specific keys ahead of time.

krunalhinguu: if we are importing specific keys, i can remove this lines. It was failing for me in local thats why i have added this

andypitcher (commented):
To get rid of this warning, change `as` to `AS`
```
 1 warning found (use --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 6) 
```
FROM --platform=$BUILDPLATFORM rancher/mirrored-tonistiigi-xx:1.3.0 `AS` xx

package/Dockerfile
andypitcher: @krunalhinguu I encountered the same error locally, and the build seems to fail also right now.
We need to find a way to import the following gpg key into the final bci-micro:15.6 image (looks like the key isn't present by default).
```
pub   rsa4096/0xF74F09BC3FA1D6CE 2023-01-19 [SC] [expires: 2027-01-18]
Key fingerprint = 7F00 9157 B127 B994 D5CF  BE76 F74F 09BC 3FA1 D6CE
uid           SUSE Package Signing Key <build@suse.de>
```
See this link: https://documentation.suse.com/sles/15-SP5/html/SLES-all/cha-update-preparation.html#sec-update-preparation-update

I'm looking into this further.

andypitcher: @krunalhinguu I found a solution to fix this issue.

The blocker in our case originates from the golang image which is used as the builder (registry.suse.com/bci/golang:1.22 AS builder), and is build on top of `15-SP5`. 

```
cat /etc/*release
NAME="SLES"
VERSION="15-SP5"
VERSION_ID="15.5"
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP5"
ID="sles"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:15:sp5"
DOCUMENTATION_URL="https://documentation.suse.com/"
```

So when the keys are downloaded it uses https://updates.suse.com/SUSE/Products/SLE-BCI/**15-SP5**/x86_64/product/repodata/repomd.xml.key.

To get over this, the value of `$releasever` can be overwritten in zypper using the `--releasever` global option, so that we can fetch the corresponding repodata for 15-SP6. In our case we can then add `--releasever=15.6` during the zypper refresh.

```
diff --git a/package/Dockerfile b/package/Dockerfile
index b2aedfe..3551bba 100644
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,9 +1,9 @@
 # Platform specific images which must support all target architectures.
-FROM registry.suse.com/bci/bci-micro:15.5 AS final
+FROM registry.suse.com/bci/bci-micro:15.6 AS final
 FROM registry.suse.com/bci/golang:1.22 AS golang

 # Builder and xx only need to support the host architecture.
-FROM --platform=$BUILDPLATFORM rancher/mirrored-tonistiigi-xx:1.3.0 as xx
+FROM --platform=$BUILDPLATFORM rancher/mirrored-tonistiigi-xx:1.3.0 AS xx
 FROM --platform=$BUILDPLATFORM registry.suse.com/bci/golang:1.22 AS builder

 # Bring xx supporting tools to the builder layer, but as it is at / it
@@ -39,7 +39,7 @@ COPY --from=golang /etc/zypp/ /chroot/etc/zypp/
 # OS binaries to run kube-bench audit commands.
 # Before removing dependencies here, ensure they are not needed by cis-operator,
 # security-scan or any of the existing kube-bench tests.
-RUN zypper --non-interactive refresh && \
+RUN zypper --non-interactive --releasever=15.6 refresh && \
     zypper --installroot /chroot -n in --no-recommends findutils tar jq gawk diffutils procps systemd gzip curl && \
     zypper --installroot /chroot clean -a && \
     rm -rf /chroot/var/cache/zypp/* /chroot/var/log/zypp/* /chroot/etc/zypp
```

I haven't found the equivalent golang image that holds `SLE BCI 15-SP6`, so I propose to keep this fix for the time being, since it complies with the level of security (avoiding `--gpg-auto-import-keys`). We'll bump the golang image as soon as there's a version with `SLE BCI 15-SP6`.

@pjbgf WDYT ?

andypitcher: FYI I pushed the change through https://github.com/rancher/security-scan/pull/213/commits/af29336dd6cb4fddcc9ddc06c1bf3277105eb0fd, all checks have passed.

krunalhinguu: Thanks @andypitcher.

krunalhinguu: so these PR is good to go once we get approvals?

pjbgf (approved):

PR #225, opened 2024-07-30
build: Clone and build sonobuoy

pjbgf:
The upstream project hasn't been released in almost one year, which caused it to accumulate several stdlib Golang vulnerabilities. Once the release cadence upstream improves, this can be reverted.

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -10,6 +10,11 @@ FROM --platform=$BUILDPLATFORM registry.suse.com/bci/golang:1.22 AS builder
 # won't be copied to the final image.
 COPY --from=xx / /
 
+ARG SONOBUOY_VERSION
+# The upstream project hasn't been released in almost one year, which
+# caused it to accumulate several stdlib Golang vulnerabilities.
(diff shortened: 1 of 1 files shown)


package/Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM registry.suse.com/bci/golang:1.22 AS golang
 
 # Builder and xx only need to support the host architecture.
 FROM --platform=$BUILDPLATFORM rancher/mirrored-tonistiigi-xx:1.3.0 AS xx
-FROM --platform=$BUILDPLATFORM registry.suse.com/bci/golang:1.22 AS builder
+FROM --platform=$BUILDPLATFORM golang AS builder
```
macedogm, line 7: Out of curiosity, why do we need to change from BCI?

pjbgf: I didn't, this was redundant. It is referring to the line `3` above:
```
FROM registry.suse.com/bci/golang:1.22 AS golang
```

The PR review not showing it makes it confusing.

macedogm: Ohh, I see. Thanks for clarifying.

pjbgf: However, it turns out that forces cross emulation instead of cross compilation, so I will remove it.

macedogm (approved):

PR #239, opened 2024-08-20
fix condition for etcd node detection for k3s

vardhaman22:
issue: https://github.com/rancher/rancher/issues/46663
In https://github.com/rancher/security-scan/pull/218 the check for etcd node for k3s was updated to `if [[ "$(ps -e | grep etcd | wc -l)" -gt 0 ]] || [[ "$(journalctl -D $JOURNAL_LOG -u k3s -u k3s-agent | grep -m1 "etcd cluster" | wc -l )" -gt 0 ]];`
this condition detects controlplane node also as an etcd node because running 
`journalctl -u k3s -u k3s-agent | grep -m1 "etcd cluster"`
produces output
`.....k3s[1548]: time="2024-08-20T07:31:30Z" level=info msg="Managed etcd cluster not yet initialized"`
because of that etcd checks are running on control plane nodes as well and failing which is resulting in mixed state for 2.x checks. 
To fix i have added an OR condition in the earlier check with the log which was mentioned in PR https://github.com/rancher/security-scan/pull/218.

### DIFF
--- a/package/run_sonobuoy_plugin.sh
+++ b/package/run_sonobuoy_plugin.sh
@@ -64,7 +64,7 @@ mkdir -p "${RESULTS_DIR}"
 
 # etcd
 if [[ "${OVERRIDE_BENCHMARK_VERSION}" != "" ]]; then
-  if [[ "$(ps -e | grep etcd | wc -l)" -gt 0  ]]  || [[ "$(journalctl -D $JOURNAL_LOG -u k3s -u k3s-agent | grep -m1 "etcd cluster" | wc -l )" -gt 0 ]]; then
+  if [[ "$(ps -e | grep etcd | wc -l)" -gt 0  ]]  || [[ "$(journalctl -m -u k3s | grep -m1 "Managed etcd cluster initializing" | wc -l )" -gt 0 ]] || [[ "$(journalctl -m -u k3s | grep -m1 "Managed etcd cluster bootstrap already complete and initialized" | wc -l )" -gt 0 ]]; then
     echo "etcd: Using OVERRIDE_BENCHMARK_VERSION=${OVERRIDE_BENCHMARK_VERSION}"
     kube-bench run \
       --targets etcd \


package/run_sonobuoy_plugin.sh
```diff
@@ -64,7 +64,7 @@ mkdir -p "${RESULTS_DIR}"
 
 # etcd
 if [[ "${OVERRIDE_BENCHMARK_VERSION}" != "" ]]; then
-  if [[ "$(ps -e | grep etcd | wc -l)" -gt 0  ]]  || [[ "$(journalctl -D $JOURNAL_LOG -u k3s -u k3s-agent | grep -m1 "etcd cluster" | wc -l )" -gt 0 ]]; then
+  if [[ "$(ps -e | grep etcd | wc -l)" -gt 0  ]]  || [[ "$(journalctl -D $JOURNAL_LOG -u k3s -u k3s-agent | grep -m1 "Managed etcd cluster initializing" | wc -l )" -gt 0 ]] || [[ "$(journalctl -D $JOURNAL_LOG -u k3s -u k3s-agent | grep -m1 "Managed etcd cluster bootstrap already complete and initialized" | wc -l )" -gt 0 ]]; then
```
dereknola, line 67:
```suggestion
  if [[ "$(ps -e | grep etcd | wc -l)" -gt 0  ]]  || [[ "$(journalctl -D $JOURNAL_LOG -u k3s -u k3s-agent | grep -m1 "Managed etcd cluster" | wc -l )" -gt 0 ]]; then
```

This was my original suggestion that did not get implemented in https://github.com/rancher/security-scan/pull/218

dereknola (changes requested):

package/run_sonobuoy_plugin.sh
dereknola, line 67:
```suggestion
  if [[ "$(ps -e | grep etcd | wc -l)" -gt 0  ]]  || [[ "$(journalctl -m -u k3s | grep -m1 "Managed etcd cluster initializing" | wc -l )" -gt 0 ]] || [[ "$(journalctl -m -u k3s | grep -m1 "Managed etcd cluster bootstrap already complete and initialized" | wc -l )" -gt 0 ]]; then
```

dereknola: Two nits:
1) using -m eliminates the issues around finding the correct JOURNAL_LOG location (this was resolved in scans https://github.com/rancher/security-scan/pull/237)
2) k3s-agent will never run etcd.

dereknola (changes requested):

package/run_sonobuoy_plugin.sh
vardhaman22: updated the command as per the suggestion

andypitcher (approved):
LGTM, thanks.

dereknola (approved):

PR #241, opened 2024-08-23
move etcd checks from master to etcd.yaml for rke2 profiles

vardhaman22:
issue: https://github.com/rancher/cis-operator/issues/350

checks 1.1.7, 1.1.8, 1.1.11 and 1.1.12 are related to etcd configuration. In upstream profiles these checks resides in master.yaml. The issue is that master.yaml checks are run on the nodes only where apiserver is detected as per
the condition
https://github.com/rancher/security-scan/blob/4299f432990ce90d1132a7b50084d6932434d586/package/run_sonobuoy_plugin.sh#L98-L103
 in run_sonobuoy_script . So the above mentioned checks fails in case etcd is deployed on a separate node.
These checks will work as expected if controlplane and etcd both are on the same node.

There is a related issue posted on upstream as well. And in a comment ( https://github.com/aquasecurity/kube-bench/issues/306#issuecomment-500135745 ) on that issue, There are two suggestions given one to move these checks in separate section and the other one is to mark them skip.
So It is  decided that we move these in etcd.yaml for the our rke2 profiles.
As per  the condition https://github.com/rancher/security-scan/blob/4299f432990ce90d1132a7b50084d6932434d586/package/run_sonobuoy_plugin.sh#L65-L70 in run_sonobuoy_plugin script the etcd.yaml checks will run on the nodes on which etcd is detected. 
So even if a cluster has single master node (controlplane + etcd) or separate controlplane and etcd nodes, the etcd.yaml checks will run on any type of node where etcd is detected. So it works in both cases.

- removed 1.1.7, 1.1.8 checks from master.yaml from rke2 profiles.
- added 1.1.7, 1.1.8, 1.1.11, 1.1.12 checks to etcd.yaml for rke2 profiles.

### DIFF
--- a/package/cfg/rke2-cis-1.23-hardened/etcd.yaml
+++ b/package/cfg/rke2-cis-1.23-hardened/etcd.yaml
@@ -7,6 +7,11 @@ type: "etcd"
 groups:
   - id: 1.1
     text: "Master Node Configuration Files"
+    # These checks (1.1.7,1.1.8,1.1.11,1.1.12) are related to etcd and in upstream they are present
+    # in master.yaml, But these checks fail on controlplane nodes which does not have etcd.
+    # so these are moved to etcd.yaml, so that the scan passes on the clusters which have seperate
+    # controlplane, etcd node. These checks will also run on controlplane nodes which have etcd since
+    # run_sonobuoy_plugin.sh script detects etcd process and runs on any node where etcd is found.
     checks:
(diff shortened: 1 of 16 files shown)


dereknola (approved):

andypitcher:
As discussed, could you provide more details in this PR (and potentially commenting the checks) about why we are making this decision.

PR #243, opened 2024-08-28
Improve Master 1.2.3 DenyServiceExternalIPs for RKE/RKE2/K3s

andypitcher:
Parent: 
- https://github.com/rancher/rancher/issues/46350

Note:

- The check has been changed to Manual for all distributions, the requirement since 1.7 is to have this flag enabled. The check will be turned to `automated` for RKE2, for CIS-1.9 once the option will be added to `profile: cis` .
- `have` does not exist in kube-bench, must be changed to `has`.
image


vardhaman22 (approved):
LGTM

PR #231, opened 2024-08-05
Overhaul RKE2 1.XX Checks

dereknola:
- Fix audits and remediations for RKE2 Master checks
- Remove duplicate etcd checks in etcd section. These are just duplicates of checks in master and don't match upstream scans

PR #247, opened 2024-09-04
rke2 fix failed checks for permissive profiles

vardhaman22:
issue: https://github.com/rancher/rancher/issues/46812 and https://github.com/rancher/cis-operator/issues/350
1.1.1, 1.1.3 and 1.1.5 checks were marked as scored: false for 1.8 permissive profile in PR: https://github.com/rancher/security-scan/pull/231
Updated these checks for the other rke2 permissive profiles based on that. 
- updated all the checks in rke2-cis-1.7-permissive profile as they were same as rke2-cis-1.8-permissive profile
- updated 1.1.3 and 1.1.5 (required permission 600) only in rke2-cis-1.24-permissive profile since for 1.1.1 the requirement is of permissions 644
- did not update  rke2-cis-1.23-permissive profile since required permission for all the checks are 644.

### DIFF
--- a/package/cfg/rke2-cis-1.24-permissive/master.yaml
+++ b/package/cfg/rke2-cis-1.24-permissive/master.yaml
@@ -40,7 +40,7 @@ groups:
         scored: true
 
       - id: 1.1.3
-        text: "Ensure that the controller manager pod specification file permissions are set to 600 or more restrictive (Automated)"
+        text: "Ensure that the controller manager pod specification file permissions are set to 600 or more restrictive (Manual)"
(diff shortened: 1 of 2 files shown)


dereknola (changes requested):
These checks pass in my testing. Are you setting `profile: cis` in the RKE2 config?

vardhaman22:
> These checks pass in my testing. Are you setting `profile: cis` in the RKE2 config?
The profile cis option was not used. We run permissive profile checks on non hardened clusters. these checks were earlier documented as expected failures. But the above mentioned [PR](https://github.com/rancher/security-scan/pull/231) set them to `scored: false` for 1.8 profile i have updated the other permissive profiles as well to keep them in sync.

andypitcher (approved):

dereknola (approved):

PR #251, opened 2024-10-11
K3s cis-1.9 profile

dereknola:
## Changes
- Add new cis-1.9 profile for K3s, specifically K3s 1.27 minor (as cis-1.10 now covers 1.28-1.31)

   This is largely a copy paste from cis-1.8, the major changes being in the policy section, with the use of `kubectl` to validate the 5.1.1-5.1.6. Note that K3s is only "Automated" for 2 of them, the rest still require manual intervention, but there is an actual audit. 

   Screenshot from 2024-10-11 10-30-28

   Screenshot from 2024-10-11 10-29-50

- Reduce log output of validate-yaml by default. Currently this scripts generates about 65K of log lines when running, effectively nullifying any actionable output (see [here](https://github.com/rancher/security-scan/actions/runs/11296889675/job/31422641481?pr=251) as example). Instead, we now only setup `set -x` if the DEBUG variable is set. This produces only about 50 lines of logs

### DIFF
--- a/package/cfg/k3s-cis-1.9/controlplane.yaml
+++ b/package/cfg/k3s-cis-1.9/controlplane.yaml
@@ -1,0 +1,62 @@
+---
+controls:
+version: "cis-1.9"
+id: 3
+text: "Control Plane Configuration"
+type: "controlplane"
+groups:
+  - id: 3.1
+    text: "Authentication and Authorization"
+    checks:
+      - id: 3.1.1
+        text: "Client certificate authentication should not be used for users (Manual)"
+        type: "manual"
+        remediation: |
+          Alternative mechanisms provided by Kubernetes such as the use of OIDC should be
+          implemented in place of client certificates.
+        scored: false
+
+      - id: 3.1.2
+        text: "Service account token authentication should not be used for users (Manual)"
+        type: "manual"
+        remediation: |
+          Alternative mechanisms provided by Kubernetes such as the use of OIDC should be implemented
+          in place of service account tokens.
+        scored: false
+
+      - id: 3.1.3
+        text: "Bootstrap token authentication should not be used for users (Manual)"
+        type: "manual"
+        remediation: |
+          Alternative mechanisms provided by Kubernetes such as the use of OIDC should be implemented
+          in place of bootstrap tokens.
+        scored: false
+
+  - id: 3.2
+    text: "Logging"
+    checks:
+      - id: 3.2.1
+        text: "Ensure that a minimal audit policy is created (Manual)"
+        audit: "journalctl -m -u k3s | grep 'Running kube-apiserver' | tail -n1 | grep 'audit-policy-file'"
+        tests:
+          test_items:
+            - flag: "--audit-policy-file"
+              set: true
+        remediation: |
+          Create an audit policy file for your cluster.
+        scored: false
+
+      - id: 3.2.2
+        text: "Ensure that the audit policy covers key security concerns (Manual)"
+        type: "manual"
+        remediation: |
+          Review the audit policy provided for the cluster and ensure that it covers
+          at least the following areas,
+          - Access to Secrets managed by the cluster. Care should be taken to only
+            log Metadata for requests to Secrets, ConfigMaps, and TokenReviews, in
+            order to avoid risk of logging sensitive data.
+          - Modification of Pod and Deployment objects.
+          - Use of `pods/exec`, `pods/portforward`, `pods/proxy` and `services/proxy`.
+          For most requests, minimally logging at the Metadata level is recommended
+          (the most basic level of logging).
+        scored: false
--- a/package/cfg/k3s-cis-1.9/policies.yaml
+++ b/package/cfg/k3s-cis-1.9/policies.yaml
@@ -1,0 +1,384 @@
+---
+controls:
+version: "cis-1.9"
+id: 5
+text: "Kubernetes Policies"
+type: "policies"
+groups:
+  - id: 5.1
+    text: "RBAC and Service Accounts"
+    checks:
+      - id: 5.1.1
+        text: "Ensure that the cluster-admin role is only used where required (Manual)"
+        audit: |
+          kubectl get clusterrolebindings -o=custom-columns=NAME:.metadata.name,ROLE:.roleRef.name,SUBJECT:.subjects[*].name --no-headers | while read -r role_name role_binding subject
+          do
+            if [[ "${role_name}" != "cluster-admin" && "${role_binding}" == "cluster-admin" ]]; then
+              is_compliant="false"
+            else
+              is_compliant="true"
+            fi;
+            echo "**role_name: ${role_name} role_binding: ${role_binding} subject: ${subject} is_compliant: ${is_compliant}"
+          done
+        use_multiple_values: true
+        tests:
+          test_items:
+            - flag: "is_compliant"
+              compare:
+                op: eq
+                value: true
+        remediation: |
+          Identify all clusterrolebindings to the cluster-admin role. Check if they are used and
+          if they need this role or if they could use a role with fewer privileges.
+          Where possible, first bind users to a lower privileged role and then remove the
+          clusterrolebinding to the cluster-admin role :
+          kubectl delete clusterrolebinding [name]
+        scored: false
+
+      - id: 5.1.2
+        text: "Minimize access to secrets (Automated)"
+        audit: "echo \"canGetListWatchSecretsAsSystemAuthenticated: $(kubectl auth can-i get,list,watch secrets --all-namespaces --as=system:authenticated)\""
+        tests:
+          test_items:
+            - flag: "canGetListWatchSecretsAsSystemAuthenticated"
+              compare:
+                op: eq
+                value: no
+        remediation: |
+          Where possible, remove get, list and watch access to Secret objects in the cluster.
+        scored: true
+
+      - id: 5.1.3
+        text: "Minimize wildcard use in Roles and ClusterRoles (Manual)"
+        audit: |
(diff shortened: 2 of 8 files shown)


package/cfg/k3s-cis-1.9/policies.yaml
```diff
@@ -0,0 +1,376 @@
...
+            - flag: "cluster-admin"
+              compare:
+                op: valid_elements
+                value: cluster-admin, helm-kube-system-traefik, helm-kube-system-traefik-crd
+        remediation: |
+          Identify all clusterrolebindings to the cluster-admin role. Check if they are used and
+          if they need this role or if they could use a role with fewer privileges.
+          Where possible, first bind users to a lower privileged role and then remove the
+          clusterrolebinding to the cluster-admin role :
+          kubectl delete clusterrolebinding [name]
```
andypitcher, line 27: Nice rework, would you mind to adapt the remediation so that we explain quickly why the  `helm-kube-system-traefik`, `helm-kube-system-traefik-crd` get whitelisted in k3s ?

package/cfg/k3s-cis-1.9/policies.yaml
```diff
@@ -0,0 +1,380 @@
...
+            if echo "${clusterrole_rules}" | grep -q "\[\"\*\"\]"; then
+              echo "**clusterrole_name: ${clusterrole_name} clusterrole_rules: ${clusterrole_rules} clusterrole_is_compliant: false"
+            else
+              printf "**clusterrole_name: %-50s clusterrole_is_compliant: true\n" "${clusterrole_name}"
+            fi;
+          done
+        remediation: |
+          Where possible replace any use of wildcards in clusterroles and roles with specific
+          objects or actions.
+        scored: false
```
andypitcher, line 74: Are there any Roles and ClusterRoles with "*" in K3s that are required, and that we could whitelist too ? (Same as 5.1.1 approach).

package/cfg/k3s-cis-1.9/policies.yaml
```diff
@@ -0,0 +1,380 @@
...
+                value: "true"
+              set: true
+        remediation: |
+          Create explicit service accounts wherever a Kubernetes workload requires specific access
+          to the Kubernetes API server.
+          Modify the configuration of each default service account to include this value
+          automountServiceAccountToken: false
+        scored: false
+
+      - id: 5.1.6
```
andypitcher, line 107: Same here, any Service Account Tokens that we could whitelist ?

package/cfg/k3s-cis-1.9/policies.yaml
dereknola: I am unable to completely fix this check due to limitations on how kube-bench processes it checks. The `or` operation will not work, on the failure detection for the first operator, the process is restarted **ON ALL LINES** for the second operation. This effectively makes every operation an implicit and.

Example given the following lines of output:
```
clusterrole_name: admin clusterrole_is_compliant: true
clusterrole_name: clustercidrs-node clusterrole_is_compliant: true
clusterrole_name: edit clusterrole_is_compliant: true
clusterrole_name: k3s-cloud-controller-manager clusterrole_is_compliant: false
```
Assume a check of `is_compliant: true` OR `clusterrole_name: k3s-cloud-controller-manager`:

The first check for `is_compliant` will fail on the 4th line. This causes the process to move onto the next check "is the clusterrole_name == k3s-cloud-controller-manger" but it restarts on the top line, not on the 4th line that failed. So the second check will fail on the first line, because `admin != k3s-cloud-controller-manager`.

package/cfg/k3s-cis-1.9/policies.yaml
dereknola: I have reworked 5.1.6 to cover all k3s whitelist exceptions. With this check we could potentially change this to `scored: true`.

andypitcher: Great, let's change `5.1.6` to `scored: true` then, as you reworked the check to `PASS` for default K3s. `5.1.6` has also been automated (`scored: true`) as of CIS-1.9.

image

package/cfg/k3s-cis-1.9/policies.yaml
andypitcher: From what I recall, it was one of the issue I stumbled upon when writing the CIS-1.9 checks. I initially had split this check into two distinct checks, but then had to follow the CIS check numbering, as advised by kube-bench's folks.

In this situation, since we can't break the check down into 2 checks, we could either:
-  improve kube-bench engine OR
- improve the bash script, to only use `is_compliant` for both `roles` and `clusterroles`. Similar to 5.1.6, and use the valid_elements as you did.

See below:

```
     - id: 5.1.3
         text: "Minimize wildcard use in Roles and ClusterRoles (Manual)"
         audit: |
           # Check Roles
           kubectl get roles --all-namespaces -o custom-columns=ROLE_NAMESPACE:.metadata.namespace,ROLE_NAME:.metadata.name --no-headers | while read -r role_namespace role_name
           do
             role_rules=$(kubectl get role -n "${role_namespace}" "${role_name}" -o=json | jq -c '.rules')
             if echo "${role_rules}" | grep -q "\[\"\*\"\]"; then
               printf "**role_name: %-50s  role_namespace: %-25s role_rules: %s is_compliant: false\n" "${role_name}" "${role_namespace}" "${role_rules}"
             else
               printf "**role_name: %-50s role_namespace: %-25s is_compliant: true\n" "${role_name}" "${role_namespace}"
             fi;
           done
           # Check ClusterRoles
           kubectl get clusterroles -o custom-columns=CLUSTERROLE_NAME:.metadata.name --no-headers | while read -r clusterrole_name
           do
             clusterrole_rules=$(kubectl get clusterrole "${clusterrole_name}" -o=json | jq -c '.rules')
             if echo "${clusterrole_rules}" | grep -q "\[\"\*\"\]"; then
               echo "**clusterrole_name: ${clusterrole_name} clusterrole_rules: ${clusterrole_rules}  is_compliant: false"
             else
               printf "**clusterrole_name: %-50s is_compliant: true\n" "${clusterrole_name}"
             fi;
           done
         use_multiple_values: true
         tests:
           bin_op: or
           test_items:
             - flag: "is_compliant"
               compare:
                 op: eq
                 value: true
             - flag: "service_account"
               compare:
                 op: valid_elements
                 value: ...,... <- put the roles and cluster roles to whitelist
```

WDYT ?

package/cfg/k3s-cis-1.9/policies.yaml
```diff
@@ -45,48 +45,50 @@ groups:
...
         remediation: |
           Where possible replace any use of wildcards in clusterroles and roles with specific objects or actions.
-          K3s requires exceptions for following cluster roles, whic are required for regular operations:
-          - k3s-cloud-controller-manager, local-path-provisioner-role
-        scored: false
+          K3s gives exceptions for following cluster roles, which are required for regular operations:
+          - k3s-cloud-controller-manager, local-path-provisioner-role, cluster-admin
+          - system:kube-controller-manager, system:kubelet-api-admin, system:controller:namespace-controller, 
+          - system:controller:disruption-controller, system:controller:generic-garbage-collector, 
+          - system:controller:horizontal-pod-autoscaler, system:controller:resourcequota-controller
```
andypitcher: Do we want to specify the nature (what are their usage) and justification for each cluster role that are whitelisted ? Maybe this could be later documented into the K3s Docs and a link to it can be added there.

package/cfg/k3s-cis-1.9/policies.yaml
```diff
@@ -45,48 +45,50 @@ groups:
...
             else
-              printf "**role_name: %-50s role_namespace: %-25s role_is_compliant: true\n" "${role_name}" "${role_namespace}"
+              printf "**role_name: %-50s role_namespace: %-25s is_compliant: true\n" "${role_name}" "${role_namespace}"
             fi;
           done
 
+          cr_whitelist="cluster-admin k3s-cloud-controller-manager local-path-provisioner-role"
+          cr_whitelist="$cr_whitelist system:kube-controller-manager system:kubelet-api-admin system:controller:namespace-controller"
+          cr_whitelist="$cr_whitelist system:controller:disruption-controller system:controller:generic-garbage-collector"
+          cr_whitelist="$cr_whitelist system:controller:horizontal-pod-autoscaler system:controller:resourcequota-controller"
```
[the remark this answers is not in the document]
dereknola: I'm fine with whatever format, so I'll change it to a array

andypitcher (dismissed):
LGTM, thanks.

package/cfg/k3s-cis-1.9/controlplane.yaml
```diff
@@ -0,0 +1,62 @@
...
+        remediation: |
+          Alternative mechanisms provided by Kubernetes such as the use of OIDC should be implemented
+          in place of bootstrap tokens.
+        scored: false
+
+  - id: 3.2
+    text: "Logging"
+    checks:
+      - id: 3.2.1
+        text: "Ensure that a minimal audit policy is created (Manual)"
```
pjbgf, line 39: This has "(Manual)", should it not have `type: manual`?

pjbgf (dismissed):

dereknola:
I'm still seeing errors on with this latest commit, so hold off on merging this.

pjbgf (dismissed):

dereknola:
We can't use a nice bash array, as the audits are run under `sh`. So a string concatenated list is the best we have.

dereknola:
Okay, its now ready, ready. here is an Updated result of the test 
image

image

pjbgf (approved):

andypitcher (approved):

PR #256, opened 2024-10-15
[v0.3] bump kube-bench

vardhaman22:
- updated kube bench to 0.9.0


andypitcher (changes requested):
Could you apply the same as https://github.com/rancher/security-scan/pull/258#pullrequestreview-2371793144. Thanks

vardhaman22:
> Could you apply the same as [#258 (review)](https://github.com/rancher/security-scan/pull/258#pullrequestreview-2371793144). Thanks

since gke 1.6.0 only supports k8s versions >=1.29 i have removed it for v2.8 rancher since max supported k8s version on rancher is v1.28.x

andypitcher (approved):

PR #257, opened 2024-10-15
added gke-1.6.0

vardhaman22:
issue: https://github.com/rancher/rancher/issues/47551
- added gke-1.6.0
- PRs for kube bench bump needs to be merged for this as well. https://github.com/rancher/security-scan/pull/253 and https://github.com/rancher/security-scan/pull/252
**Note: gke 1.2.0 is not removed since gke 1.6.0 is only for versions >=1.29 and other gke profiles are not available in upstream kube-bench and the minimum supported k8s version in rancher v2.8 is 1.25**

### DIFF
--- a/hack/e2e
+++ b/hack/e2e
@@ -60,7 +60,7 @@ function check_binaries(){
 
 function check_config_files(){
   echo "> Check for upstream test files:"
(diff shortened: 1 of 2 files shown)


andypitcher (changes requested):
Could you apply the same as https://github.com/rancher/security-scan/pull/258#pullrequestreview-2371793144, too. thanks

andypitcher (approved):

PR #300, opened 2024-11-25
Add cis-1.9 generic and update configmap

andypitcher:
This PR does the following:

- Add [cis-1.9](https://github.com/aquasecurity/kube-bench/tree/main/cfg/cis-1.9) generic profile
- Add cis (1.23 to 1.9) generic `version_mappings` to configmap
- Remove old profiles and their corresponding version_mappings in configmap
- Remove generic cis versions in package/cfg since they are downloaded from upstream


package/cfg/config.yaml
```diff
@@ -211,25 +211,13 @@ managedservices:
...
-    - "rke-cis-1.20-permissive"
-  "v1.18.10+rke2r1":
-    - "rke2-cis-1.5-hardened"
-    - "rke2-cis-1.5-permissive"
+  "1.23": "cis-1.23"
+  "1.24": "cis-1.24"
+  "1.25": "cis-1.7"
+  "1.26": "cis-1.8"
+  "1.27": "cis-1.9"
+  "1.28": "cis-1.9"
```
pjbgf, line 219: Should this not go under `# CIS - Generic profiles`?

andypitcher: Thanks @pjbgf ! This part concerns the `version_mapping`, which tells kube-bench to use a specific version for a given k8s version, if no profile argument is passed.
We reorganized the target_mapping by k8s distros and profile types a while ago as it was very messy, and it's where we commented with `# CIS - Generic profiles`,  `# rke1: Generic`, etc.
IMO we could improve the comments for the version_mapping par to mention what this section is used for.

pjbgf: Thanks for clarifying, that makes sense. :bow:

pjbgf (approved):

vardhaman22 (approved):

PR #304, opened 2024-11-28
[v0.5] Add cis-1.9 generic and update configmap

andypitcher:
Inspired from #300, the difference remains on some of the `target_versions` that were left to keep the functionality.


pjbgf (approved):
LGTM. Due to lifecycle state of `v0.3`, I would refrain from updating it with this changes.

vardhaman22 (approved):

PR #325, opened 2025-01-03
[main] make 5.1.2 and 5.1.4 checks manual

vardhaman22:
because of the upstream kubectl issue:  
https://github.com/kubernetes/kubernetes/issues/93474 (duplicate: https://github.com/kubernetes/kubernetes/issues/104532)
these checks can not be automated so changed them to manual.

### DIFF
--- a/package/cfg/k3s-cis-1.9/policies.yaml
+++ b/package/cfg/k3s-cis-1.9/policies.yaml
@@ -32,6 +32,7 @@ groups:
         scored: true
 
       - id: 5.1.2
+        type: manual
         text: "Minimize access to secrets (Automated)"
         audit: "echo \"canGetListWatchSecretsAsSystemAuthenticated: $(kubectl auth can-i get,list,watch secrets --all-namespaces --as=system:authenticated)\""
         tests:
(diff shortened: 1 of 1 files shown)


andypitcher (commented):
@vardhaman22 I'll let @dereknola give his thoughts, but I'm thinking that we could use the `type: manual` and leave the check `automated`. The result will be manual, and override the automated status. Here is an [example](https://github.com/aquasecurity/kube-bench/blob/df48da449c7e5d7d4e95f605d93fa5ba2719706b/cfg/cis-1.7/policies.yaml#L180).
The reason behind this is that, we can automate (As it's stated in CIS Workbench), however the audit is limited.

dereknola (approved):

vardhaman22:
> @vardhaman22 I'll let @dereknola give his thoughts, but I'm thinking that we could use the `type: manual` and leave the check `automated`. The result will be manual, and override the automated status. Here is an [example](https://github.com/aquasecurity/kube-bench/blob/df48da449c7e5d7d4e95f605d93fa5ba2719706b/cfg/cis-1.7/policies.yaml#L180). The reason behind this is that, we can automate (As it's stated in CIS Workbench), however the audit is limited.

updated the PR with type: manual

pjbgf (approved):

PR #447, opened 2025-03-19
Add K3s cis-1.10 benchmark

dereknola:
Adds k3s-cis-1.10 benchmark.
Same as upstream [cis-1.10 new additions](https://github.com/aquasecurity/kube-bench/commit/3a2348eba7ff4168206b08e59e462be20c27ce72#diff-5de7e26c7dcf850edd8a6e922a4815d27f52bc45f92068ac6837b50fcc23dcd4) with 1 whitelist exception:
- 5.2.9 Requires an exception for coredns and svclb traefik as they require additional capability as a matter of regular K3s operations. (Both have additional network capability)

### DIFF
--- a/package/cfg/k3s-cis-1.10/policies.yaml
+++ b/package/cfg/k3s-cis-1.10/policies.yaml
@@ -1,0 +1,576 @@
+---
+controls:
+version: "cis-1.10"
+id: 5
+text: "Kubernetes Policies"
+type: "policies"
+groups:
+  - id: 5.1
+    text: "RBAC and Service Accounts"
+    checks:
+      - id: 5.1.1
+        text: "Ensure that the cluster-admin role is only used where required (Automated)"
+        audit: |
+          kubectl get clusterrolebindings -o=custom-columns=ROLE:.roleRef.name,NAME:.metadata.name,SUBJECT:.subjects[*].name --no-headers |  grep cluster-admin
+        use_multiple_values: true
+        tests:
+          test_items:
+            - flag: "cluster-admin"
+              compare:
+                op: valid_elements
+                value: cluster-admin, helm-kube-system-traefik, helm-kube-system-traefik-crd
+        remediation: |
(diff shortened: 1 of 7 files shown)


package/cfg/k3s-cis-1.9/policies.yaml
```diff
@@ -107,7 +107,7 @@ groups:
         scored: true
 
       - id: 5.1.5
-        text: "Ensure that default service accounts are not actively used. (Automated)"
+        text: "Ensure that default service accounts are not actively used. (Manual)"
```
andypitcher, line 110:
```suggestion
        text: "Ensure that default service accounts are not actively used. (Automated)"
```
I would lean towards keeping this `automated` as we can audit K3s in an automated way, then add whitelisting if needed. This will likely make the check to `FAIL` instead of `WARN`, but then as I you stated, the user can manually fix by following the doc.

package/cfg/k3s-cis-1.10/policies.yaml
```diff
@@ -455,6 +464,8 @@ groups:
           Audit: the audit retrieves each Pod's container(s) added capabilities.
           Condition: is_compliant is false if added capabilities are added for a given container.
           Default: Containers run with a default set of capabilities as assigned by the Container Runtime.
+          K3s gives exceptions to the following pods, which are required for regular operations:
```
andypitcher, line 467: Agreed, would be good to develop on why `coredns` and `svclb-traefik` are whitelisted in the K3s doc.

package/cfg/k3s-cis-1.9/policies.yaml
dereknola: Removed that change from PR

andypitcher (dismissed):

vardhaman22 (dismissed):
LGTM
overall similar to k3s cis 1.9
just some audits are added for 5.2.2-6 and 5.2.9 with scored: false

andypitcher (approved):

PR #536, opened 2025-10-06
[v0.6] Update release creation step 

krunalhinguu:
We are updating the 'Create Release' task in the workflow. The current script fails due to a dirty Git state, and resolving it would require testing multiple solutions. Instead, we are replacing it with an already verified and implemented release workflow.


.github/workflows/release.yml
```diff
@@ -65,13 +65,7 @@ jobs:
           prime-username: ${{ env.PRIME_REGISTRY_USERNAME }}
           prime-password: ${{ env.PRIME_REGISTRY_PASSWORD }}
       - name: Create release
-        if: matrix.tag-suffix == ''   # only run for the main image
+        run: |
+          gh release create ${{ github.ref_name }} --verify-tag --title "${{ github.ref_name }}" --generate-notes
```
andypitcher, line 69: Could you explain this change ?

krunalhinguu: we are changing the script as current script is failing CI

andypitcher (approved):

PR #537, opened 2025-10-06
[v0.7] Update release creation step 

krunalhinguu:
We are updating the 'Create Release' task in the workflow. The current script fails due to a dirty Git state, and resolving it would require testing multiple solutions. Instead, we are replacing it with an already verified and implemented release workflow.


andypitcher:
Could you bump kube-bench version as this commit https://github.com/rancher/security-scan/commit/2101b50f01eea5a7d5b1dcbe8b1ab776d5e390cb, e2e is failing because of that.

pjbgf (approved):

andypitcher (approved):

PR #589, opened 2026-01-27
Fix rke2 check 1.2.11, improve wording

dereknola:
We cannot utilize the  `use_multiple_values` modifier this because that requires all values to be the same result. 

Tested locally on a `profile: cis` rke2 1.32 cluster. 

https://github.com/rancher/compliance-operator/issues/168


andypitcher (approved):
LGTM as discussed offline, thanks !

Small note, we could technically change this check to an `Automated` check, as we can easily test it on any RKE2 setup (the flag will always be captured through the kube-apiserver process definition). This could take place later with other checks.

PR #601, opened 2026-02-23
[main] added aks 1.7 support

Abhishek-Valaboju:
issue: https://github.com/rancher/rancher/issues/53866

added aks 1.7 support

### DIFF
--- a/package/cfg/config.yaml
+++ b/package/cfg/config.yaml
@@ -229,8 +229,6 @@ version_mapping:
     - "gke-1.2.0"
   "gke-1.6.0":
     - "gke-1.6.0"
-  "aks-1.0":
-    - "aks-1.0"
   "aks-1.7":
     - "aks-1.7"
   "v1.24.17+k3s1": "k3s-cis-1.24-hardened"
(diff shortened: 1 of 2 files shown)


mitulshah-suse (dismissed):

package/cfg/config.yaml
```diff
@@ -258,6 +260,8 @@ target_mapping:
   # AKS
   "aks-1.0":
     - "node"
+  "aks-1.7":
+    - "node"
```
andypitcher:
```suggestion
    - "master"
    - "node"
    - "controlplane"
    - "policies"
    - "managedservices"
```
As per upstream's [configmap](https://github.com/aquasecurity/kube-bench/blob/8a7204d42f0bea44db453c545db052c29f58a9cc/cfg/config.yaml#L474) it seems that we're missing targets (even for aks-1.0).

andypitcher (changes requested):
LGTM, just added a small suggestion regarding the configmap for aks profiles, if you could adapt it.

package/cfg/config.yaml
```diff
@@ -258,6 +260,12 @@ target_mapping:
   # AKS
   "aks-1.0":
     - "node"
```
andypitcher:
```suggestion
```
same as : https://github.com/rancher/security-scan/pull/601/changes#r2852543169

andypitcher (dismissed):

andypitcher (approved):

PR #629, opened 2026-03-30
[main] CI improvements

Teja78906:
CI improvements
Added checksum verification and certificate validation for downloaded dependencies using curl, wget , go install 
upgrade golangci-lint version to v2.11.4


andypitcher (approved):
LGTM - will provide a new PR to replace the `get_k8s_api_version()` to use kubectl instead.

PR #698, opened 2026-08-24
rke2/k3s cis-2.0 scans

dereknola:
These are exact copies of the cis-1.12 scans, but with a wording change for `3.2.2`. Note that the upstream https://github.com/aquasecurity/kube-bench/pull/2126 also included a wording change to `1.2.29` but rke2/k3s use custom wording for this spec as is, so it was a no-op.


andypitcher (approved):
LGTM, thanks

